PR #1, opened 2026-05-12
Add macOS support via Python + bleak daemon

stereoabuse:
## Summary

- Adds a parallel Python + [bleak](https://github.com/hbldh/bleak) daemon for macOS since the bash daemon depends on Linux-only tools (`bluetoothctl`, `busctl`, `dbus-monitor`) that have no shell-level macOS equivalent.
- Mirrors the bash daemon 1:1: same UUIDs, cache path, poll cadence (60s + 5s tick), refresh-on-notify behaviour, JSON payload schema. The firmware doesn't change.
- Auto-start via a per-user LaunchAgent (so the macOS Bluetooth permission prompt can surface in the GUI session). `install-macos.sh` creates `daemon/.venv/`, installs bleak, and bootstraps the agent.
- Linux bash daemon, systemd unit, and `install.sh` are unchanged.

## Test plan

I don't have a Clawdmeter to test against, so the BLE behaviour needs verification on the device. What still needs to be checked:

- [ ] `./install-macos.sh` runs to completion on a fresh macOS Mac
- [ ] First-run macOS prompt to allow Python access to Bluetooth fires and accepting it lets the daemon proceed
- [ ] After pairing the device in System Settings → Bluetooth, the daemon scans by name and connects
- [ ] JSON payload reaches the firmware (Usage screen populates)
- [ ] Fresh-boot REQ notify (`0x01` on `...0004`) triggers an immediate poll
- [ ] Disconnecting and re-powering the device → daemon reconnects on its own
- [ ] `launchctl kickstart -k …` restart and `launchctl bootout …` cleanup both work
- [ ] `tail -f ~/Library/Logs/claude-usage-daemon.log` shows timestamped progress

What's already verified on my Mac (no hardware involved):

- [x] `python3 -m py_compile daemon/claude-usage-daemon.py` clean
- [x] Module loads cleanly with bleak 3.0.2 (`from bleak import BleakClient, BleakScanner`)
- [x] `build_payload()` output matches the README's documented format byte-for-byte: `{"s":45,"sr":120,"w":28,"wr":7200,"st":"allowed","ok":true}`
- [x] `plutil -lint` on the substituted plist

## Notes for review

- **Pairing UX divergence.** macOS has no `bluetoothctl pair` equivalent — users pair once via System Settings → Bluetooth. The firmware enables bonding (`setSecurityAuth(true, false, true)` in `firmware/src/ble.cpp`) but the RX characteristic is `WRITE | WRITE_NR` only (no `WRITE_ENC`), so unpaired writes may also succeed — worth checking on hardware.
- **Cache file is a CoreBluetooth UUID, not a MAC.** On macOS, bleak surfaces opaque peripheral identifiers instead of MAC addresses. The daemon caches whatever bleak returns at the same path the bash daemon uses (`~/.config/claude-usage-monitor/ble-address`). The MAC shown on the firmware's Bluetooth screen isn't directly usable on macOS, but scan-by-name handles first contact.
- **Bluetooth TCC prompt.** First BLE access by the venv's Python interpreter triggers a system permission prompt. If the user misses it, it can be granted later in System Settings → Privacy & Security → Bluetooth. Documented in the README.


stereoabuse:
Closing — opened this prematurely. Wanted to test it locally first.

PR #5, opened 2026-05-14
Add macOS host support and Waveshare 1.8" AMOLED board port

lorddavidson:
## Summary

Adds **macOS host support** and a **Waveshare ESP32-S3-Touch-AMOLED-1.8 board** port, alongside the existing 2.16" Linux target. Both pre-existing paths are preserved — nothing breaks for current users.

### macOS host (replaces `bluetoothctl` / `busctl` / `systemd`)
- `daemon/claude_usage_daemon.py` — Python rewrite of the bash daemon using `bleak` (CoreBluetooth). Mirrors the original's behavior: 60s poll, refresh-on-notify, 1→60s reconnect backoff, MAC/UUID caching.
- `daemon/com.user.claude-usage-daemon.plist` — LaunchAgent template (Mac equivalent of the systemd unit).
- `install-mac.sh` — installs to `~/Library/LaunchAgents/`, sets up a Python venv, primes Bluetooth permission interactively.
- `flash-mac.sh` — wraps `pio run -t upload` with `/dev/cu.usbmodem*` autodetect and `-e <env>` selection.
- Token reading on macOS pulls from the **Keychain** (`Claude Code-credentials` service) since `~/.claude/.credentials.json` is Linux-only; Linux path is preserved as a fallback.

### Firmware: Waveshare ESP32-S3-Touch-AMOLED-1.8 (368×448)
Gated by `-DBOARD_AMOLED_18` — original 2.16" build is untouched at runtime.

- `firmware/platformio.ini` — refactored into a `[common]` base + two envs (`waveshare_amoled_216`, `waveshare_amoled_18`).
- `firmware/src/display_cfg.h` — board-conditional pins, resolution, driver typedefs:
  - Display: **Arduino_SH8601** (vs CO5300), SCLK 11 (vs 38), reset via TCA9554 P0
  - Touch: **TouchDrvFT6X36** at 0x38 (vs CST92xx at 0x5A), INT on GPIO 21
  - New: TCA9554 I/O expander at 0x20 drives reset lines
- `firmware/src/main.cpp` — conditional object construction; minimal inline TCA9554 driver (reset pulse before display init); **`rotate_strip` rewritten to handle non-square panels** (was hardcoded `S = LCD_WIDTH` which broke on 368×448); `rot_buf` sized by `max(LCD_W, LCD_H)`.
- `firmware/src/ui_layout.h` (new) — per-board layout constants for `ui.cpp` (margins, panel sizes, splash cell size).
- `firmware/src/ui.cpp` / `firmware/src/splash.cpp` — consume `ui_layout.h`; splash canvas uses 18px cells on 1.8 (360×360 centered) vs 24px on 2.16 (480×480).

## Testing

- ✅ `pio run -e waveshare_amoled_216` builds clean (regression check — confirms refactors didn't break upstream behavior)
- ✅ `pio run -e waveshare_amoled_18` builds clean
- ✅ Flashed and ran on real ESP32-S3-Touch-AMOLED-1.8 hardware: display renders correctly, touch axes correct, splash animations centered, BLE GATT round-trips with the Mac daemon
- ✅ Mac daemon connects, polls Anthropic API, writes payload, receives `REQ` notifications
- ⚠️ **Not regression-tested on Linux** — I don't have a Linux setup. The original `daemon/claude-usage-daemon.sh`, `install.sh`, `flash.sh`, and `.service` files are unmodified, so Linux behavior should be intact, but a second pair of eyes there would help.

## Notes

- Happy to **split this into two PRs** (macOS support / 1.8 board port) if you'd prefer smaller review surface — let me know.
- The 1.8 port's touch swap/mirror and SH8601 col/row offsets are best-guesses that worked on my unit; if other 1.8 owners see issues I'll iterate.
- No upstream assets, font files, or licensing-sensitive material modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,15 @@ firmware/.vscode/
 # Node tooling deps (npm install --no-save)
 tools/node_modules/
 tools/package-lock.json
+
+# Mac daemon venv + local pio install
+daemon/.venv/
+.pio-venv/
+
+# Python bytecode cache
+__pycache__/
+*.pyc
+
+# IDEs
+.idea/
+.vscode/
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -1,0 +1,331 @@
+#!/usr/bin/env python3
+"""Claude Usage Tracker Daemon (BLE) — macOS port of claude-usage-daemon.sh.
+
+Polls Claude API rate-limit headers and writes a JSON payload to the
+ESP32 "Claude Controller" peripheral over a custom GATT service. Uses
+bleak (CoreBluetooth backend on macOS).
+"""
+
+import asyncio
+import getpass
+import json
+import os
+import re
+import signal
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+import httpx
+from bleak import BleakClient, BleakScanner
+from bleak.exc import BleakError
+
+DEVICE_NAME = "Claude Controller"
+SERVICE_UUID = "4c41555a-4465-7669-6365-000000000001"
+RX_CHAR_UUID = "4c41555a-4465-7669-6365-000000000002"
+REQ_CHAR_UUID = "4c41555a-4465-7669-6365-000000000004"
+
+POLL_INTERVAL = 60
+TICK = 5
+SCAN_TIMEOUT = 8.0
+
+# macOS: token lives in Keychain (service "Claude Code-credentials").
+# Linux: token lives in ~/.claude/.credentials.json.
+KEYCHAIN_SERVICE = "Claude Code-credentials"
+CREDENTIALS_PATH = Path.home() / ".claude" / ".credentials.json"
+SAVED_ADDR_FILE = Path.home() / ".config" / "claude-usage-monitor" / "ble-address"
+
+API_URL = "https://api.anthropic.com/v1/messages"
+API_HEADERS_TEMPLATE = {
+    "anthropic-version": "2023-06-01",
+    "anthropic-beta": "oauth-2025-04-20",
+    "Content-Type": "application/json",
+    "User-Agent": "claude-code/2.1.5",
+}
+API_BODY = {
+    "model": "claude-haiku-4-5-20251001",
+    "max_tokens": 1,
+    "messages": [{"role": "user", "content": "hi"}],
+}
+
+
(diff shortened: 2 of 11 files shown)


HermannBjorgvin:
Thanks for the PR! I'll take a look in a few hours when I get back home, I think I even have one of the 1.8" boards to test as well.

HermannBjorgvin:
🤖AI Generated: Tested this on the 2.16 board — build/flash/runtime all clean, daemon connects, Usage screen renders identically, layout constants and `rotate_strip` math both check out. Thanks for the careful refactor and the `ui_layout.h` split.

One thing I wanted to ask about before I sign off — I don't have a 1.8" board on me right now so I can't verify this myself, but reading through the diff it looks like `main.cpp`'s button loop is unchanged, while from the Waveshare docs the 1.8 only ships two physical buttons (BOOT on GPIO 0, PWR on **TCA9554 EXIO4** rather than the AXP). On paper that would mean:

- `BTN_FWD = GPIO 18` has no physical button on the 1.8, so Shift+Tab (Claude Code mode toggle) wouldn't fire.
- `power_pwr_pressed()` reads via the AXP path, which isn't how the 1.8's PWR button is wired, so cycling screens / splash animations wouldn't work either — and the Bluetooth status screen would be unreachable.

How's it actually behaving for you on the unit you flashed? Curious whether I'm reading the wiring wrong, or whether you're already navigating via touch / a different path I'm not seeing. If there is a gap there, one shape I had in mind was remapping PWR (via EXIO4) → Shift+Tab and letting a touch tap cycle \`Usage → Bluetooth → Splash\`, but I'd rather hear how it works for you today before suggesting changes.

HermannBjorgvin:
Hey @lorddavidson I had techcrunch reach out the project so expedited the macos support and cherry picked parts from your branch into #6 I hope you don't mind. 

I wanted to have the macos support as soon as possible due to that article coming out, I credited you in the macos instructions in the README.md, really appreciate your work in verifying that.

lorddavidson:
Hi @HermannBjorgvin, great news and of course I don't mind. Will look at fixing the 1.8 issues/questions over the weekend!

Thank you also for the mention :)

________________________________

HermannBjorgvin:
Closing since #25 added 1.8" device support. Please bring a PR if you have any feedback on it of course!

PR #9, opened 2026-05-15
Add dockerized PlatformIO build/flash workflow

jjhuff:
## Summary
- Add `firmware/Dockerfile` — Python 3.11-slim base with PlatformIO 6.1.19 and the OS deps it pulls in for fetching, building, and flashing (git, curl, srecord, udev).
- Add `firmware/pio-docker.sh` wrapper around `docker run` for the common workflows: `build`, `upload`, `monitor`, `clean`, `shell`, and a `--` passthrough. Builds the image on first use, caches PlatformIO state in a named volume, binds the serial device for upload/monitor.

Lets contributors build and flash without installing the ESP32 toolchain on the host.

## Testing

I tested this against some Lilygo hardware I have. Working on a port.

Thanks!

### DIFF
--- a/firmware/Dockerfile
+++ b/firmware/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.11-slim
+
+# System deps PlatformIO uses for fetching, building, and flashing.
+# udev so esptool can enumerate USB serial devices.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        git \
+        ca-certificates \
+        curl \
+        srecord \
+        udev \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
This is super cool! I guess it should work cross platform as well since you're just binding the port, I expect that works on both MacOS and Linux?

I'll take a look at this after the weekend if that's ok I got a small stack of pull requests to look at but I think this should probably be the default installation method going forward

jjhuff:
Thanks! Docker is a PITA sometimes, but I also don't like having tools installing all over my system.

It should work cross platform. I'll see if I can get a coworker to test on mac.

HermannBjorgvin:
Been feeling the need to get a Mac for that exact reason 😄

PR #10, opened 2026-05-15
Add dockerized BLE daemon

jjhuff:
## Summary
- Drop-in container for running `daemon/claude_usage_daemon.py` without installing Python deps on the host.
- Compose file mounts the host's system D-Bus so bleak reaches BlueZ, plus `~/.claude/.credentials.json` (ro) and `~/.config/claude-usage-monitor/` (rw, for the cached MAC).
- AppArmor's default container profile blocks D-Bus `AddMatch` on Ubuntu/Debian, so the compose file sets `apparmor=unconfined`.

Linux only — Docker Desktop on macOS can't reach host Bluetooth; Mac users keep using `install-mac.sh` / launchd.

## Quick start
```
docker compose -f daemon/docker-compose.yml up -d --build
docker compose -f daemon/docker-compose.yml logs -f
```

NOTE: Only one BLE client can hold the GATT connection — stop any systemd unit or foreground daemon before starting the container.

Also, I discovered the python-based daemon after I opened the other docker PR. Sorry for the bonus related PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,17 @@ Check status: `systemctl --user status claude-usage-daemon`
 
 View logs: `journalctl --user -u claude-usage-daemon -f`
 
+### Run the daemon in Docker (alternative)
+
+Skip the systemd unit and run the daemon in a container instead. The compose file mounts the host's system D-Bus so `bleak` reaches BlueZ, plus your credentials file and the cached MAC dir.
+
+```bash
+docker compose -f daemon/docker-compose.yml up -d --build
+docker compose -f daemon/docker-compose.yml logs -f
+```
+
+Only one BLE client can hold the GATT connection — stop the systemd unit (`systemctl --user stop claude-usage-daemon`) before starting the container. Docker Desktop on macOS can't reach host Bluetooth, so this path is Linux only.
+
 ## How it works
 
 1. The daemon reads your Claude Code OAuth token from `~/.claude/.credentials.json`.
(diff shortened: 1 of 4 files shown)


HermannBjorgvin:
The python daemon was specifically for the macos support, there a bash daemon available for Linux with no python dep which uses bluez

Wondering if it's worth dockerizing the daemon only for Linux since bluez and bash are the only deps there

What do you think?

jjhuff:
The bash daemon wasn't finding my credentials -- but I didn't investigate why. I'll see what I can find.

It's nice that the python daemon works cross platform. Maybe easier to add more features, like other data/stats. Docker allows it to be more isolated if needed. Perhaps easier distribution/install too. (On Linux)

The bash one is easy to audit, but didn't really dockerize well.

HermannBjorgvin:
Oh that's interesting, if it's giving other people greif as well then it might indeed be a good idea to dockerize the python one instead

jjhuff:
Ahhh, I have multiple `accessToken`s in my .credentials.json due to some mcp servers.

`jq` would make short work of this, if you wanted to keep the shell version (and don't mind the extra dependency) 
```
  read_token() {
      jq -r '.claudeAiOauth.accessToken // empty' "$HOME/.claude/.credentials.json"
  }
```

The python version seems to support a top-level `accessToken` token as well, so the solution above might need to be more complicated

PR #17, opened 2026-05-17
feat: add minimal Freenove ESP32-S3 Display NonTouch support

tungtobe:
## Summary

- Add a new `freenove_s3_display_minimal` PlatformIO environment targeting the **Freenove ESP32-S3 Display NonTouch FNK0104B** (2.8" 240×320 ILI9341)
- Keep the existing Waveshare `waveshare_amoled_216` target fully intact — this is a parallel addition, not a replacement
- Update README to document both supported targets with flashing instructions and a capability comparison table

## What's included

The Freenove target adds board-specific files alongside the existing Waveshare code:

| File | Purpose |
|------|---------|
| `firmware/src/main_freenove.cpp` | Freenove entry point (setup/loop) |
| `firmware/src/ui_freenove.h/.cpp` | Minimal single-screen usage UI |
| `firmware/src/display_freenove.h/.cpp` | ILI9341 display backend |
| `firmware/src/freenove_board.h` | Pin definitions for FNK0104B |

The minimal target keeps the core flow (BLE GATT data service, Usage screen, host daemon protocol) while intentionally omitting Waveshare-specific features not available on the Freenove board: touch input, splash animations, Bluetooth screen, HID buttons, battery UI/PMU, and IMU auto-rotation.

## Testing

- Built `waveshare_amoled_216` ✅
- Built `freenove_s3_display_minimal` ✅
- Flashed to a Freenove ESP32-S3 Display NonTouch FNK0104B ✅
- macOS daemon connected over BLE ✅
- Usage payloads rendered on device ✅

## Notes

The Freenove FNK0104B uses an ILI9341 display controller. During bring-up, using ST7789 resulted in a lit backlight with no rendered pixels; switching to the ILI9341 configuration (as documented in the Freenove wiki) resolved the issue.

---
*Tested on macOS with the Python BLE daemon from the `macos-support` PR.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,14 @@
 
 A small ESP32 dashboard I made for my desk to keep an eye on Claude Code usage.
 
-It runs on a [Waveshare ESP32-S3-Touch-AMOLED-2.16](https://www.waveshare.com/esp32-s3-touch-amoled-2.16.htm?&aff_id=149786) and pairs with my laptop over Bluetooth, the splash screen plays pixel-art Clawd animations that get
-busier when your usage rate climbs. The two side buttons send Space and
-Shift+Tab over BLE HID for Claude Code's voice mode and mode-toggle shortcuts.
+It supports two firmware targets:
+
+| Target | Board | Experience |
+| --- | --- | --- |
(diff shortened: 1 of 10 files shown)


HermannBjorgvin:
Hey thanks for your contribution! Here's a bit of an update:

I held off on merging any alternative hardware ports because I wanted a good abstraction layer for that. Then @tobby168 came in clutch and delivered that in #25 with instructions in the `docs` folder on how to implement your port.

We include detailed instructions now which hopefully would allow your clanker to refactor your PR to conform with the new standard, thank you for your patience!

P.S. if you can post a picture of your port that would be appreciated, I enjoy seeing how it looks on people's devices!

PR #6, opened 2026-05-14
Add macOS host support

HermannBjorgvin:
## Summary

Cherry-picks the macOS-specific pieces from #5 so they can land independently while the 1.8" board port iterates separately.

Credits **[Chris Davidson (@lorddavidson)](https://github.com/lorddavidson)** as the original author of all the macOS code — thanks!

## What's included

- `daemon/claude_usage_daemon.py` — Python rewrite of the bash daemon using `bleak` (CoreBluetooth on macOS). Mirrors the bash daemon's behavior: 60 s poll, refresh-on-notify, exponential reconnect backoff, MAC/UUID caching.
- `daemon/com.user.claude-usage-daemon.plist` — LaunchAgent template (Mac equivalent of the systemd unit).
- `install-mac.sh` — installs to `~/Library/LaunchAgents/`, sets up a Python venv with `bleak` + `httpx`, primes Bluetooth permission interactively.
- `flash-mac.sh` — wraps `pio run -t upload` with `/dev/cu.usbmodem*` autodetect.
- `README.md` — split into "macOS installation" and "Linux installation" sections with parallel Flash / Pair / Install subsections.

## What's not included

The 1.8" board firmware port from #5 is intentionally left out — those button-wiring questions are still being discussed on the original PR and don't block macOS support.

## What changed vs #5

- `flash-mac.sh` simplified: dropped the `-e <env>` flag and the `waveshare_amoled_18` default (no second env exists on this branch).
- `.gitignore` additions trimmed to the macOS-relevant entries (`daemon/.venv/`, `__pycache__/`, `*.pyc`).
- README rearranged so macOS and Linux are sibling H2s rather than the macOS section sitting under a Linux-flavored Prerequisites header.

## Test plan

- [x] Linux side untouched at runtime — `install.sh` / `flash.sh` / bash daemon / systemd unit unchanged
- [ ] macOS install end-to-end (flash → pair → install daemon → see usage update on device) — needs a Mac to verify

PR #18, opened 2026-05-17
Make BLE pairing work on macOS, allow daemon + HID to coexist

tobby168:
Hi @HermannBjorgvin 👋 — really love this project, such a clever little
desk gadget, thanks for building it (and for accepting #6 to open the
door for Mac users). Sharing back the fixes I needed to get it talking
to my Mac end-to-end.

## Summary

Five focused firmware/script changes that fix the macOS host pieces shipped in #6 — verified end-to-end on macOS 15.7.3 against a live Anthropic API token.

Before this PR, on a Mac:

1. The device advertises but **never appears in System Settings → Bluetooth's Nearby Devices list** — macOS Sequoia hides BLE peripherals that don't explicitly advertise the standard HID service UUID.
2. Once visible (manually or via a BLE scanner app like LightBlue), there's **no Connect button** — macOS refuses to surface pairing for a HID claiming to be Apple's Magic Keyboard (Apple's USB vendor `0x05AC` + product `0x820A`).
3. After a successful pair, the **Keyboard Setup Assistant** pops up on every reconnect — the HID descriptor doesn't include an LED output report, and the country code is `0` (Not Supported).
4. Once paired, the **daemon can no longer reach the device** — macOS holds the single available connection slot for the HID link and the daemon's `BleakScanner` never sees the peripheral.

Each fix is independent and minimal.

## Firmware changes (`firmware/src/ble.cpp`)

- **Advertise the standard HID Service UUID `0x1812` in the primary advertising packet.** The custom 128-bit data-service UUID had to move to the scan response — keeping both in the primary packet overflowed the 31-byte BLE advertising limit and macOS's passive scanner doesn't read scan responses.
- **Switch PnP ID to Espressif's BT SIG vendor (`0x02E5`)** instead of Apple's USB vendor + Magic Keyboard product. macOS validates Apple-claimed HIDs and silently denies pairing for impostors.
- **Add the LED output report** (Num/Caps/Scroll Lock) to the HID descriptor. macOS treats a keyboard descriptor without LEDs as "incomplete" and triggers the Keyboard Setup Assistant on every fresh pair.
- **Set HID country code to 33 (US ANSI)** instead of 0 (Not Supported) so macOS can identify the layout without asking the user.
- **Bump `CONFIG_BT_NIMBLE_MAX_CONNECTIONS` to 2 and restart advertising after each accept.** macOS holds one connection for the HID keyboard; the daemon now gets its own slot for the data service in parallel instead of either side starving the other.

## Helper script changes (`screenshot.sh`)

- Auto-pick `/dev/cu.usbmodem101` on macOS vs `/dev/ttyACM0` on Linux instead of hardcoding the Linux path.
- Fall back to PlatformIO's bundled Python at `~/.platformio/penv/bin/python` if pyserial isn't on the system Python (PEP 668 blocks `pip install pyserial` on Homebrew Python by default).
- Pass the actual framebuffer dimensions reported in the `SCREENSHOT_START` header to ffmpeg instead of hardcoding `480x480` — useful preparation for any future board with a different resolution.

## Daemon changes (`daemon/claude_usage_daemon.py`)

- **Log API HTTP status + response body on 4xx/5xx and return None** instead of building a payload with all-zero fields. The previous behavior silently swallowed token-expiry 401s and pushed `{"s":0,"sr":0,"w":0,"wr":0,...}` to the device, so users saw a stuck-at-zero usage display with no clue why.

## Test plan

- [x] AMOLED-2.16 firmware still builds cleanly (`pio run -d firmware -e waveshare_amoled_216`)
- [x] Fresh pair on macOS 15.7.3 — device shows up in System Settings → Bluetooth, Connect button present, Keyboard Setup Assistant skipped on subsequent reconnects (after first cancel)
- [x] BOOT button sends Space over BLE HID after pairing (verified with `python3 -c "print(input())"` while pressing)
- [x] Daemon connects via UUID after one initial scan (the OS-held HID connection no longer blocks the second BLE slot)
- [x] Daemon now logs `API HTTP 401: {"type":"error",...}` on expired tokens instead of silently sending zeroes
- [x] `./screenshot.sh out.png` works on macOS without arguments

## Known follow-up (not in this PR)

The daemon could refresh expired access tokens itself using the `refreshToken` already stored in the same keychain blob (`claudeAiOauth.refreshToken`). For now the user has to `claude /login` again when the token expires; happy to do this in a follow-up if there's interest.

### DIFF
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -166,6 +166,9 @@ async def poll_api(token: str) -> dict | None:
     except httpx.HTTPError as e:
         log(f"API call failed: {e}")
         return None
+    if resp.status_code >= 400:
+        log(f"API HTTP {resp.status_code}: {resp.text[:200]}")
+        return None
 
     def hdr(name: str, default: str = "0") -> str:
         return resp.headers.get(name, default)
--- a/firmware/platformio.ini
+++ b/firmware/platformio.ini
@@ -11,12 +11,14 @@ build_flags =
     -DBOARD_HAS_PSRAM
     ; AXP2101 PMU
     -DXPOWERS_CHIP_AXP2101
-    ; NimBLE config — peripheral-only, single connection
+    ; NimBLE config — peripheral, 2 simultaneous connections so the OS can
+    ; hold the HID link (Space key from BOOT button) while the daemon holds
+    ; its own connection for the custom data service.
     -DCONFIG_BT_NIMBLE_ROLE_BROADCASTER=1
     -DCONFIG_BT_NIMBLE_ROLE_PERIPHERAL=1
     -DCONFIG_BT_NIMBLE_ROLE_CENTRAL=0
     -DCONFIG_BT_NIMBLE_ROLE_OBSERVER=0
-    -DCONFIG_BT_NIMBLE_MAX_CONNECTIONS=1
+    -DCONFIG_BT_NIMBLE_MAX_CONNECTIONS=2
     ; LVGL config via build flags
     -DLV_CONF_SKIP
     -DLV_COLOR_DEPTH=16
--- a/firmware/src/ble.cpp
+++ b/firmware/src/ble.cpp
@@ -13,7 +13,10 @@
 
 #define BLE_BUF_SIZE 512
 
-// HID keyboard report descriptor
+// HID keyboard report descriptor (standard 6-KRO boot-protocol-compatible).
+// Includes the LED output report (Num/Caps/Scroll Lock indicators) — without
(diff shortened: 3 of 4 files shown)


HermannBjorgvin:
Thank you so much for those fixes!

Looks good to me so I think I'll just merge as is

HermannBjorgvin:
Regarding the token refresh not sure if the daemon should handle it or if we let Claude Code take care of that, we're piggybacking off of the Claude Code credentials.json so I'm guessing they'll handle that for us but maybe for systems with a long uptime we could have a daemon with an expired token in that case maybe we read the credentials.json file again until we see a new token?

PR #19, opened 2026-05-18
docs: link to ccusage-mqtt for the no-ESP32 path

george-vice:
Hey @HermannBjorgvin — really nice project. I built [ccusage-mqtt](https://github.com/george-vice/ccusage-mqtt), an MQTT publisher that mirrors Clawdmeter's telemetry surface for Home Assistant. The `mood`/burn-rate thresholds are ported verbatim from `firmware/src/usage_rate.cpp` with a golden test against the firmware behavior, so calibration stays in sync with whatever Clawdmeter does.

It's aimed at folks who want the dashboard but don't want to put an ESP32 on their desk — they get a Home Assistant device with 15 sensors + a pre-built Lovelace card instead.

Added a one-line entry under a "Related projects" section. Happy to adjust the wording / placement if you'd rather it sit somewhere else, or close this if it doesn't fit. Thanks for the firmware — the calibration ported cleanly. 🙏

- Repo: https://github.com/george-vice/ccusage-mqtt
- Landing: https://george-vice.github.io/ccusage-mqtt/

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,10 @@ See `tools/README.md` for details.
 - Lucide icon set ([lucide.dev](https://lucide.dev), MIT) for bluetooth and battery UI glyphs.
 - Anthropic brand fonts (Tiempos Text, Styrene B) — see licensing warning below.
 
+## Related projects
+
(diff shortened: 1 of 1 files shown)


HermannBjorgvin:
If it had any interoperability I could see the reasoning but otherwise too many similar projects for a related section in the readme, sorry.

PR #21, opened 2026-05-18
Keep BLE UI state at CONNECTED while re-advertising

georgetasioulis:
## Summary
After #18 enabled `MAX_CONNECTIONS=2`, `onConnect` sets `need_advertise=true` so a second central (the host daemon) can grab a slot alongside the OS-held HID link. `start_advertising()` then unconditionally set `state = BLE_STATE_ADVERTISING`, so with one connected client the UI flipped to "Advertising…" on the next `ble_tick()` and never recovered until a second client attached.

Only update the state to `ADVERTISING` when `getConnectedCount() == 0`.

## Test plan
- [x] macOS Sequoia, single client: pair from Bluetooth Settings, verify UI flips to "Connected" and stays there.
- [ ] Two clients (OS HID + host daemon): UI remains "Connected" with both slots in use.
- [x] Disconnect: UI flips back to "Advertising…" once the last client leaves.

### DIFF
--- a/firmware/src/ble.cpp
+++ b/firmware/src/ble.cpp
@@ -87,8 +87,16 @@ static void start_advertising() {
     adv->setScanResponseData(scanResp);
     adv->enableScanResponse(true);
     bool ok = adv->start();
-    state = BLE_STATE_ADVERTISING;
-    Serial.printf("BLE: advertising start=%s\n", ok ? "OK" : "FAILED");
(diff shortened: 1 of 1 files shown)


HermannBjorgvin (approved):

HermannBjorgvin:
Thanks @georgetasioulis ! 

Really happy to see people creating pull requests instead of issues 😃

PR #24, opened 2026-05-19
Auto-sleep screen after 30 min idle

tobby168:
## Summary

After being left on overnight the device drains the battery because the AMOLED stays lit. This adds a simple idle-screen-off so it can sit on the desk indefinitely.

- AMOLED brightness fades to 0 after `IDLE_TIMEOUT_MS` (default 30 min) with no physical button activity.
- Any of the three physical buttons (`BTN_BACK` / `BTN_FWD` / PWR) fades it back in.
- The **first press from sleep is consumed for wake only** — Space / Shift+Tab / cycle-screen only fire on the second press, so grabbing the device to look at it doesn't accidentally send PTT to Claude or jump screens.
- **Touch never counts as activity** and is fully swallowed while asleep: pets/sleeves can't wake it, and LVGL can't quietly toggle splash↔usage on a dark panel.
- Rotation handling defers while asleep so its blank+ramp doesn't fight the idle fade; a rotation that happened during sleep gets ramped in on wake.

AMOLED `brightness=0` is true 0 emission on this panel, so no extra power-management plumbing (light-sleep, AXP rail gating) is needed to get the battery savings — the new code is purely a small state machine plus a few hook sites in `main.cpp`.

All tunables live in the new `idle_cfg.h` (timeout, fade durations, default brightness) — nothing hard-coded in `main.cpp`. The previously hard-coded `gfx->setBrightness(200)` at init and the trailing `200` in the rotation ramp now both flow from `DISPLAY_DEFAULT_BRIGHTNESS`.

## Files

- `firmware/src/idle_cfg.h` — new, configuration
- `firmware/src/idle.h` / `firmware/src/idle.cpp` — new, state machine (`AWAKE → FADING_OUT → ASLEEP → FADING_IN`) with linear-interpolated `setBrightness()` fades
- `firmware/src/main.cpp` — `idle_init()` replaces the literal init brightness; `idle_tick()` runs once per loop; touch read gated; the three button handlers route their press edges through `idle_consume_wake_press()`; rotation handler skips while asleep

## Test plan

- [x] Builds clean with `pio run -d firmware -e waveshare_amoled_216`
- [x] Boots; splash animation visible at default brightness
- [x] Verified on hardware with `IDLE_TIMEOUT_MS` temporarily lowered to 15 s:
  - Panel fades to black after timeout
  - PWR fades it back in **without** cycling to the next screen; a second PWR press cycles normally
  - `BTN_BACK` wakes the panel **without** sending HID Space to the host; a second press sends Space
  - Touching the dark panel does not wake it and does not toggle splash↔usage
- [x] Restored `IDLE_TIMEOUT_MS` to 30 min before commit

### DIFF
--- a/firmware/src/idle.cpp
+++ b/firmware/src/idle.cpp
@@ -2,6 +2,7 @@
 #include "idle.h"
 #include "idle_cfg.h"
 #include "display_cfg.h"   // declares `extern Arduino_CO5300 *gfx;`
+#include "power.h"
 
 enum IdleState {
     STATE_AWAKE,
@@ -68,6 +69,16 @@ bool idle_is_asleep(void) {
 
 void idle_tick(void) {
     uint32_t now = millis();
+
+    // While on USB power (if configured), don't sleep — and wake from sleep
+    // when power comes back. Treats USB-in as continuous activity.
+    if (!IDLE_SLEEP_WHEN_CHARGING && power_is_vbus_in()) {
+        last_activity_ms = now;
+        if (state == STATE_ASLEEP || state == STATE_FADING_OUT) {
+            begin_fade(DISPLAY_DEFAULT_BRIGHTNESS, now);
+            state = STATE_FADING_IN;
+        }
+    }
 
     switch (state) {
     case STATE_AWAKE:
--- a/firmware/src/idle.cpp
+++ b/firmware/src/idle.cpp
@@ -1,0 +1,103 @@
+#include <Arduino.h>
+#include "idle.h"
+#include "idle_cfg.h"
+#include "display_cfg.h"   // declares `extern Arduino_CO5300 *gfx;`
+
+enum IdleState {
+    STATE_AWAKE,
+    STATE_FADING_OUT,
+    STATE_ASLEEP,
+    STATE_FADING_IN,
+};
+
+static IdleState state = STATE_AWAKE;
+static uint32_t last_activity_ms = 0;
+static uint32_t fade_started_ms  = 0;
+static uint32_t fade_last_step_ms = 0;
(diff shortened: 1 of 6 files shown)


tobby168:
@HermannBjorgvin another useful change port from my branch, no rush please take time to help review, thanks!

HermannBjorgvin:
Maybe suggestion to allow touch to wake from sleep? What do you think?

Also a suggestion maybe to not sleep when plugged in? I like looking at the little Clawd animations on my desk, not sure what the burn-in risk is but so worth it imo.

Some people also have some alternative hardware that doesn't have a battery so maybe odd for them to include a sleep functionality.

tobby168:
Thanks for the review! Pushed e0829ce addressing all three points with two new flags in `idle_cfg.h`:

- **`IDLE_SLEEP_WHEN_CHARGING`** (default `false`) — while USB is plugged in we don't sleep and we wake immediately if power comes back. Covers both your "I want to watch Clawd on my plugged-in desk" and the no-battery hardware case in one go.
- **`IDLE_WAKE_ON_TOUCH`** (default `true`) — touch on the dark panel wakes, with the first touch swallowed (same wake-consumption pattern as the buttons, so tapping a dark screen doesn't also flip splash↔usage). Set `false` to keep the pets/sleeves-safe behaviour.

To make the "plugged in" detection work correctly on battery-less builds I also added `power_is_vbus_in()` in `power.{h,cpp}` — it uses `pmu.isVbusIn()` rather than `isCharging()`, so the "plugged in, nothing to charge" case is detected.

Let me know what you think!

HermannBjorgvin:
Thanks! That's awesome to have!

PR #28, opened 2026-05-22
macos uses keychain; remove stale comments

BrianPugh:
on macos, `~/.claude/.credentials.json` doesn't exist, so i replaced that warning in `install-mac.sh`


HermannBjorgvin (approved):

HermannBjorgvin:
Thanks for the fix!

PR #29, opened 2026-05-22
switch daemon to `/api/oauth/usage` endpoint

BrianPugh:
More proper than using the dummy "hi" message query.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ View logs: `journalctl --user -u claude-usage-daemon -f`
 ## How it works
 
 1. The daemon reads your Claude Code OAuth token — from the macOS Keychain (service `Claude Code-credentials`) on macOS, or from `~/.claude/.credentials.json` on Linux.
-2. It makes a minimal API call to `api.anthropic.com/v1/messages` — one token of Haiku, basically free.
-3. The usage numbers come straight out of the response headers (`anthropic-ratelimit-unified-5h-utilization` and friends).
+2. It GETs `api.anthropic.com/api/oauth/usage` (the same endpoint `claude /usage` uses) — costs zero tokens.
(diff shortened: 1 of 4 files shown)


HermannBjorgvin:
Had a coworker try that endpoint but he ran into some rate limiting with that endpoint, did you have any issues with that?

BrianPugh:
I've been running it at a 60 second refresh rate for the past few days without issue, but I haven't tried it at a higher refresh rate.

HermannBjorgvin:
Fair enough, I think that's a good change then, doesn't start people's usage sessions as well.

Sorry I merged another board port just now so I see there's a conflict on the README.md, I'll merge this when that's resolved

BrianPugh:
I just rebased on `main`, should be good to go!

BrianPugh:
@HermannBjorgvin hey sorry, actually maybe revert this merge for a little bit; scrolling through my log I did notice some rate-limiting errors. I just didn't notice because my display still updated at reasonable rates for my workflow. I think it's best to revert until we better understand the rate-limiting problem.

BrianPugh:
So i think what is/was happening is that if these any non-successfully query, it would re-attempt every 5 seconds, which would self-perpetuate a rate-limiting error. I'm adding exponential backoff with a limit of 5 minutes in-between calls. I'll let this run over the next few days and see if that resolves the issue.

HermannBjorgvin:
Thanks for the heads up, I reverted the PR but I still think this is a good idea to switch to using the API.

Let me know what you discover, thanks for the contributions!

BrianPugh:
ha, so it seems other [people thought 5min exponential backup](https://github.com/anthropics/claude-code/issues/31637) would have been a reasonable solution, but the API still limits them.

Still digging to see if there is a more proper solution...

PR #30, opened 2026-05-22
Bugfix: specify board target

BrianPugh:
Currently, the flash scripts flash ALL environements, which is clearly not what we want. The flash scripts now take in a mandatory argument specifying which argument to flash (e.g. waveshare_amoled_216).


HermannBjorgvin (approved):

HermannBjorgvin:
Thanks for the fix!

PR #33, opened 2026-05-22
Idle chime: 3-note tone on session running→idle (1.8" only)

tobby168:
Follow-up from #22 (second item under \"Known follow-ups\"). When any Claude Code session transitions `running → idle`, the device plays two short beeps through the onboard speaker. Inspired by Open Island's per-session chime UX.

## Scope

**AMOLED-1.8 only.** The 2.16 has no audio path so it gets no-op stubs; future ports opt in via a new `BoardCaps.has_audio` flag + a board-side `audio.cpp` that implements the audio HAL. Zero RAM/flash overhead on boards without audio.

## Architecture

Four layers, all behind the `has_audio` cap gate:

1. **Audio HAL** (`firmware/src/hal/audio_hal.h`) — three calls: `audio_hal_init()`, `audio_hal_play_chime()` (fire-and-forget, coalescing), `audio_hal_set_muted(bool)`. Shared code calls these unconditionally after gating on `board_caps().has_audio`.
2. **1.8 audio driver** (`boards/waveshare_amoled_18/audio.cpp`) — minimal inline ES8311 register init for 16 kHz mono playback, I2S0 via Arduino Core 3.x `ESP_I2S.h`, PA gated on only during playback. Chime synthesized procedurally (sine + linear attack/release envelope to avoid clicks) into a heap buffer, handed to I2S DMA in one write from a dedicated low-priority FreeRTOS task pinned to core 1. Pin map and codec init cross-checked against Waveshare's `examples/15_ES8311/15_ES8311.ino` for this SKU.
3. **Daemon transition detection** (`daemon/claude_usage_daemon.py`) — module-level `_previous_phase` dict tracks per-session phase across ticks. On any running→idle transition, the next payload carries a top-level `\"c\": 1`. Hoisted out of `sessions` so the chime trigger survives the BLE-budget trim loop even if every session has to be dropped. First-sighting sessions never chime — daemon restart doesn't spuriously fire per already-idle session.
4. **Mute UX** — long-press PWR (>1s, new `power_hal_pwr_long_pressed()`) toggles a NVS-persisted mute flag. Persistent \"MUTE\" tag on the Activity title row + a 1.5s centered \"Muted\" / \"Chime on\" toast on toggle. Unmuting plays a confirmation chime. Default is **unmuted** — chimes ring out of the box, hold PWR for ~1s to silence.

## Chime sound

**Two G6 beeps (1568 Hz)**, 180 ms tone + 70 ms gap = ~430 ms total. Iterated live on the 1.8\" board — a 3-note C5/E5/G5 arpeggio sounded like one blur through the small speaker; even at 160 ms/note the user perceived 2 distinct notes at most. Two clean G6 beeps land as a recognizable \"ding-ding\" that reads as \"your agent stopped.\" All chime params (frequencies, durations, envelope) live as `constexpr` at the top of `audio.cpp` — one-line edits to retune.

## Files

- `firmware/src/hal/audio_hal.h` — new HAL interface
- `firmware/src/hal/board_caps.h` — new `has_audio` field
- `firmware/src/hal/power_hal.h` — new `power_hal_pwr_long_pressed()`
- `firmware/src/boards/{waveshare_amoled_18,waveshare_amoled_216,template}/audio.cpp` — implementations (full / stub / stub)
- `firmware/src/boards/{waveshare_amoled_18,waveshare_amoled_216,template}/board.h` — new `BOARD_HAS_AUDIO` flag + I2S/codec pins on 1.8
- `firmware/src/boards/{waveshare_amoled_18,waveshare_amoled_216,template}/caps.cpp` — populate `has_audio`
- `firmware/src/boards/{waveshare_amoled_18,waveshare_amoled_216,template}/power.cpp` — long-press state machine on 1.8, no-op stubs elsewhere
- `firmware/src/data.h` — top-level `chime` flag in `UsageData`
- `firmware/src/main.cpp` — `audio_hal_init()` in setup, NVS load/save, long-press handler, chime trigger after BLE payload parse, new `chime` serial command for debug
- `firmware/src/ui.{h,cpp}` — \"MUTE\" indicator widget + toast (plain text — our custom Tiempos/Styrene fonts don't carry LV_SYMBOL_* glyphs)
- `daemon/claude_usage_daemon.py` — `_previous_phase` tracking + `consume_idle_transition()` + top-level `c` emission

## Test plan

- [x] `pio run -e waveshare_amoled_18` builds clean (flash 21.1%, RAM 35.4%)
- [x] `pio run -e waveshare_amoled_216` builds clean (verifies stubs satisfy the HAL contract)
- [x] Daemon parses (`python3 -c 'import ast; ast.parse(...)'`)
- [x] **Hardware (1.8):** flashed; `Audio HAL ready` in boot log; ES8311 init succeeds on first try with no I2C errors
- [x] **Hardware (1.8):** chime audible at default volume; two G6 beeps land cleanly
- [x] **Hardware (1.8):** long-press PWR (≥1s) toggles mute, \"MUTE\" indicator + toast render correctly, short-press still cycles screens (no double-fire)
- [x] **Hardware (1.8):** mute toggle persists across full USB power-cycle (NVS namespace `\"clawd\"`, key `\"muted\"` — auto-created on first boot to silence the previous `nvs_open failed: NOT_FOUND` log)
- [x] **Hardware (1.8):** real daemon → real Claude Code session → Stop hook → daemon emits top-level `c: 1` → firmware plays chime within ~5s of session ending

## Known gotchas / things to verify

- **PA pin discrepancy.** `IOX_PIN_PA_EN` on XCA9554 EXIO2 is held HIGH by `io_expander_init()` as it has been since the 1.8 port landed. The new code also drives `PA_CTRL_GPIO = 46` per Waveshare's reference demo. Both paths are asserted so the amp enables regardless of which the board actually routes — worth grepping the schematic before V2 if you care. (Chime is clearly audible with both lines on, so practically a no-op.)
- **Codec init values are conservative defaults** taken from the espressif/esp-adf `es8311.c` driver. Working well at DAC volume 0xBF (-32 dB) + amplitude 0.65; raise REG32 if you want it louder.
- **LV_SYMBOL_MUTE / VOLUME_MAX glyphs aren't in our fonts.** Used plain text (\"MUTE\", \"Muted\", \"Chime on\") instead. Compiling in `lv_font_montserrat_*` would unlock proper symbols if you want them in a follow-up — adds ~10KB flash.

## Note on PR base

This branch is based on PR #22's head + a merge of `activity-stream` so the new code can call into the Stop hook + per-session phase tracking that #22 introduces. **Until #22 merges, this PR's diff includes #22's commits as noise.** After #22 merges, the diff against `activity-stream` collapses to just the chime work. Happy to rebase when ready.


tobby168:
Hey @HermannBjorgvin — when you have a moment, would love a review on this one too. It's the "idle sound on Stop" follow-up I mentioned in #22. Tested end-to-end on my 1.8" board (`Audio HAL ready` on boot, real `claude -p ...` session → Stop hook → chime — see the test plan in the description for the full list).

Two notes:
- Currently 1.8"-only since that's the board with the codec + speaker; 2.16 gets no-op stubs and zero overhead. Easy to extend later if your board grows an audio path.
- The diff includes #22's commits as noise until #22 merges into `activity-stream` — happy to rebase once that lands so the diff against `activity-stream` collapses to just the chime work.

No rush on either of these — whenever the timing works for you.

HermannBjorgvin:
Wondering about the linux bash daemon and installer, wouldn't this also need to be mirrored there?

HermannBjorgvin:
Sorry I was going to test this on my end, haven't gotten around to it yet, my apologies. Will test during the work week hope that's ok

PR #34, opened 2026-05-23
Add Waveshare ESP32-S3-Touch-AMOLED-2.06 board port

samestrin:
## Summary

Adds a board port for the [Waveshare ESP32-S3-Touch-AMOLED-2.06](https://www.waveshare.com/esp32-s3-touch-amoled-2.06.htm) — a 410×502 watch-form-factor AMOLED kit. The port is a hybrid of the existing ports:

- **CO5300 display + AXP PKEY for PWR** (like the 2.16 port)
- **FT3168 touch + fixed orientation** (like the 1.8 port)
- **No IO expander** — LCD_RESET and TP_RESET are direct GPIOs pulsed in `board_init()`
- **32 MB flash** with `default_32MB.csv` partition table

Pin assignments were taken from Waveshare's reference `Mylibrary/pin_config.h` in their [demo repo](https://github.com/waveshareteam/ESP32-S3-Touch-AMOLED-2.06).

## CO5300 col_offset gotcha (worth highlighting)

The 2.06 panel exposes its visible 410-wide viewport at a horizontal offset inside the CO5300's internal RAM. Without passing the right `col_offset1` to the `Arduino_CO5300` constructor, framebuffer writes land at the wrong columns and a vertical strip of stale boot content shows through on the right edge. Waveshare's library uses `col_offset1 = 22`; this port uses `23` after fine-tuning for centering on one unit. The 2.16" port doesn't have this issue because its 480×480 viewport fills the controller's RAM. I've added a note to `docs/porting/adding-a-board.md` so the next CO5300 porter doesn't lose an afternoon to this.

## Rebased onto current main

This branch was originally cut before the sound HAL, the hold-to-pair gesture, and the BLE supervision-timeout fix landed. It's now rebased onto current `main`, which required three additions to the port beyond the original hardware bring-up:

- **`sound.cpp`** — `main.cpp` now calls `sound_hal_init/tick/play_reset` unconditionally, so the port wouldn't link without it. The 2.06 does have an ES8311 codec on the shared I2C bus, but the amp path is unverified on this board, so this no-ops (same posture as the C6 1.8 port). The shared `chime.cpp` engine is ready to wire up once someone tests it on this hardware.
- **`power.cpp`** — added `power_hal_pwr_long_pressed()` / `power_hal_pwr_released()` for hold-to-pair, mirroring the 2.16's AXP PKEY implementation (LONG + POSITIVE IRQs, plus the 8s press-off bump).
- **`platformio.ini`** — picked up the PPCP connection-parameter flags from #106 so this board doesn't ship with the Windows supervision-timeout churn.

An earlier revision of this PR also tweaked `flash-mac.sh` to take an optional env argument. #30 fixed that same footgun more strictly on `main`, so that change is dropped — this PR no longer touches `flash-mac.sh`.

## Test plan

Verified on hardware (pre-rebase build):

- [x] Display renders correctly with `col_offset1 = 23` — splash animation centered, no edge artifacts
- [x] AXP2101 PMU init OK, battery percent reads correctly
- [x] QMI8658 IMU init OK (rotation disabled — watch enclosure is fixed orientation)
- [x] FT3168 touch — basic events fire (tap-to-toggle splash) and coordinates are accurate (BLE reset zone hit-area works)
- [x] BLE pairs as "Clawdmeter", daemon (macOS) connects and pushes usage data successfully

Verified post-rebase:

- [x] `pio run -e waveshare_amoled_206` builds clean (1.34 MB, 10.2% of the app partition, 31.6% RAM)
- [x] All four existing envs still build (216, 18, 216_c6, 18_c6)

Not yet exercised on hardware:

- [ ] **Hold-to-pair on the 2.06.** The gesture postdates the original bring-up, so the PKEY LONG/POSITIVE IRQ path hasn't been run on this board. The AXP2101 is the same part at the same address as the 2.16 and the button wires straight to PWRON, so I'd expect it to behave identically — but flagging it rather than implying it was tested. Happy to confirm once I have the device in front of me again.

Photo of the working device available on request.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,12 +6,13 @@ selected via PlatformIO's `build_src_filter`. Adding a board means dropping in
 a new folder + a new `[env:...]` block — `main.cpp`, `ui.cpp`, and `splash.cpp`
 never see board-specific code. See [`docs/porting/adding-a-board.md`](docs/porting/adding-a-board.md).
 
-Four ports today (two SoC families, two panel sizes):
+Five ports today (two SoC families, three panel sizes):
 
 - `boards/waveshare_amoled_216/` — original Waveshare ESP32-S3-Touch-AMOLED-2.16 (CO5300, 480×480 square, CST9220 touch, IMU rotation). Build env: `waveshare_amoled_216`.
 - `boards/waveshare_amoled_18/` — Waveshare ESP32-S3-Touch-AMOLED-1.8 (368×448 portrait, XCA9554 IO expander). Build env: `waveshare_amoled_18`. **Two panel revisions are auto-detected at boot** (`board_rev()` in `board_init.cpp`, enum in `board_rev.h`): original = SH8601 display + FT3168 touch (0x38); later = CO5300 display + CST816 touch (0x15). One binary drives both.
 - `boards/waveshare_amoled_216_c6/` — Waveshare ESP32-C6-Touch-AMOLED-2.16 (SH8601, 480×480, CST9217 touch). Build env: `waveshare_amoled_216_c6`. ESP32-C6 SoC: single-core RISC-V, **no PSRAM**, BLE 5 only.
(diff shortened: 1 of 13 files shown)


HermannBjorgvin:
Do you have a picture of it running on the device?

HermannBjorgvin (approved):

HermannBjorgvin:
Thanks for the port! I'll merge this once the conflicts are resolved, looks good to me!

samestrin:
Waveshare ESP32-S3 Touch AMOLED 2.06

PR #39, opened 2026-05-26
Oauth usage endpoint

BrianPugh:
Revamp of #37. Changes from #37:
1. Add exponential backoff up to 5min period. This solves the previous "retry every 5 seconds" bug/bad-behavior.
2. Add a new optional `err` field to the BLE payload. If supplied, the string will be displayed instead of `anim_messages`.

So here are my findings on the rate-limiting behavior:
1.  After not using ClaudeCode for serveral hours, your authentication credentials expire and you begin getting 401 Auth errors.
3. After several authentication errors (even with 5min period), you will get 429 rate-limit-errors.
4. After you send a ClaudeCode message, ClaudeCode will renew your credentials, and the 401/429 Errors go away and everything works expected.

So ultimately, in real workflows, the rate-limiting-errors aren't really an issue. It would be nice to somehow detect when we send our first message in a while to instantly update usage (instead of up to 5min of stale data for the initial successful usage poll).

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,25 @@ JSON payload format (written to RX):
 { "s": 45, "sr": 120, "w": 28, "wr": 7200, "st": "allowed", "ok": true }
 ```
 
-Fields: `s` = session %, `sr` = session reset (minutes), `w` = weekly %, `wr` = weekly reset (minutes), `st` = status, `ok` = success flag.
+On failure the daemon instead writes a short error payload, which the firmware surfaces in place of the rotating caption on the usage screen:
+
+```json
+{ "ok": false, "err": "Rate limited" }
+```
+
+Field keys are kept short to fit in a single BLE write (ATT MTU is tight):
+
+| Key   | Type    | Meaning                                          | Range / values                                            |
+| ----- | ------- | ------------------------------------------------ | --------------------------------------------------------- |
+| `s`   | int     | Session utilization (5-hour window)              | `0`–`100` (percent)                                       |
+| `sr`  | int     | Minutes until the session window resets          | `0`+ minutes; `-1` if unknown                             |
+| `w`   | int     | Weekly utilization (7-day window)                | `0`–`100` (percent)                                       |
+| `wr`  | int     | Minutes until the weekly window resets           | `0`+ minutes; `-1` if unknown                             |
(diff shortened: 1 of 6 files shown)


HermannBjorgvin:
Me and @tobby168 had some similar discussions in #32 

I think maybe it is a related topic, since credentials expiring was also a concern there.

There is one quirk about using Claude Code to refresh the credentials and that is that it starts a new 5 hour session for the user which they might not want since a lot of people including me and my coworker have specific times when it works best for us to start a new 5 hour window.

Not sure what the best solution is here to be honest. Is the 5 minute window too broad or is it fine?

Should we leverage hooks maybe to see a session start event and start fetching usage data at that point for x number of hours?

BrianPugh:
> Not sure what the best solution is here to be honest. Is the 5 minute window too broad or is it fine?

For me, the initial up-to-5-minute-delay is totally fine. I'm not really concerned about my usage within the first 5 minutes of my session 😛 . That said, it would be nicer to have a more reactive solution. 

> Should we leverage hooks maybe to see a session start event and start fetching usage data at that point for x number of hours?

I think this would be great. Maybe use [watchfiles](https://pypi.org/project/watchfiles/) or something? I imagine ClaudeCode updates some file in the `~/.claude` directory whenever actually running.

HermannBjorgvin:
Yeah wondering what the best non-invasive way to do this is.

Perhaps changing the polling to ~5 minute intervals is best, right now it's polling every minute and storing the usage stats for the past 5 polls and using that to calculate the rate and displaying different splash screens based on that.

I like this way better indeed, can you tell me how does the splash animations work now did you update the rate calculations to account for the new time intervals?

BrianPugh:
>Perhaps changing the polling to ~5 minute intervals is best

I worry this would be too sluggish, especially for Pro ($20) users. I think the 1min interval is fine, no actual limiting happens during real sessions.

> did you update the rate calculations to account for the new time intervals?

I didn't touch any of that calculation since the normal operating rate is still the default 60s

niradler:
what about this https://github.com/anthropics/claude-desktop-buddy/blob/main/REFERENCE.md ?

HermannBjorgvin:
> what about this https://github.com/anthropics/claude-desktop-buddy/blob/main/REFERENCE.md ?

Would work for Claude desktop users, doesn't give the 5h/7d window in the data that's sent over from what I can tell... maybe they could add it for us?

PR #40, opened 2026-05-27
Add WiFi/HTTP transport variant (alongside USB CDC)

dsk2k:
Adds a third host↔board transport option — WiFi/HTTP — alongside main's BLE and @dilbery's [PR #26](https://github.com/HermannBjorgvin/Clawdmeter/pull/26) USB CDC. Daemon publishes an HTTP bridge on the host LAN, board polls it over WiFi. Compile-time selectable via build flag so existing setups don't change unless they opt in.

## Tested on real hardware

**Board**: Waveshare ESP32-S3-Touch-AMOLED-2.16 (480×480).
**Host**: Windows 10 (Dutch locale), Python 3.14, PlatformIO 6.1.19.
**Network**: standard home WiFi (2.4GHz), board → AP → host on the same /24.

Verified end-to-end:

- Daemon publishes payload at `http://<host>:8787/payload` with bearer-token auth (token persists in `~/.clawdmeter/wifi_token`, mode 0600 on POSIX). `/health` is open for diagnostics.
- Firmware connects to WiFi STA, polls the bridge every 5s (configurable `WIFI_POLL_MS`), parses the same JSON schema as USB CDC, renders Usage + Activity exactly as on the other transports.
- **`netstat` audit on host**: 24 concurrent `TimeWait` entries from the board's IP on port 8787, ~5 new unique source ports every 25s. Matches `WIFI_POLL_MS = 5000` exactly.
- **Physical USB-unplug test**: USB-C cable pulled from the host, board moved to a wall-charger. Daemon log shows the USB CDC writer recycling its port and falling through to `run_session_no_port()` (HTTP-only mode). Board continues receiving Usage + Activity live, polling over WiFi only.

## Scope of this PR

Built on top of @dilbery's [PR #26](https://github.com/HermannBjorgvin/Clawdmeter/pull/26) (USB CDC transport) plus my Windows-port and Activity backport from those discussions, so this branch carries some commits that aren't WiFi-specific. **The WiFi-only delta is the single commit `36d97d0`** — 9 files, 423 insertions:

- `firmware/src/wifi_link.{h,cpp}` (new) — public API mirrors `serial_link.h`, body wrapped in `#ifdef CLAWDMETER_TRANSPORT_WIFI` so USB-only builds don't need `wifi_config.h`.
- `firmware/src/wifi_config.h.example` — template, real `wifi_config.h` is `.gitignore`d.
- `firmware/src/main.cpp` — macro layer (`link_init` → `serial_link_init` or `wifi_link_init`) so the rest of main.cpp doesn't change between transports.
- `firmware/src/ui.h` — conditional include for the active transport's `link_state_t`.
- `firmware/platformio.ini` — new env `waveshare_amoled_216_wifi` extends the USB env + `-DCLAWDMETER_TRANSPORT_WIFI`.
- `daemon/claude_usage_daemon.py` — stdlib `http.server.ThreadingHTTPServer` thread; `_http_payload_lock`-guarded cache; helper `run_session_no_port()` keeps API + state.json polling alive when there's no USB port.
- `daemon/run-daemon.ps1` — sets `CLAWDMETER_HTTP=1` for Windows scheduled-task runs.
- `.gitignore` — excludes `wifi_config.h`.

## Design notes

- **Decoupling**: `update_http_payload()` runs unconditionally during payload assembly, so a USB CDC write failure doesn't take HTTP down (and vice versa). Both transports serve the same data without doubling the upstream API hit rate.
- **Token handling**: stdlib `secrets.token_urlsafe(32)` at first daemon start, `secrets.compare_digest()` for constant-time validation. POSIX `chmod 0o600` on the token file; on Windows the default user-profile ACL is sufficient.
- **No HTTPS** — bridge is LAN-only, same threat model as @dilbery's USB-only path. Don't expose to the Internet without a reverse proxy.
- **Architecture precedent**: same shape as [rootedlab-code/claude-code-usage-monitor](https://github.com/rootedlab-code/claude-code-usage-monitor) (different host-side data source, same transport pattern). Their bearer-token + LAN HTTP + 5s poll approach worked, ours follows.

## Known follow-ups

- **First-time WiFi setup is compile-time only.** rootedlab-code uses a captive portal on softAP (board boots as `ClaudeMonitor-XXYY` until configured). Doable here too — Arduino WiFi has the primitives — but kept out of this PR to keep the diff focused. Hardcoded `wifi_config.h` works for personal-use scenarios.
- **Daemon serves both transports simultaneously**, which produces benign `Write failed; recycling port` log spam when the board is on WiFi-only. A future cleanup is a `CLAWDMETER_TRANSPORT={USB,WIFI,BOTH}` env-var split; for now both modes coexist.

## How to use

```bash
# Host side
export CLAWDMETER_HTTP=1               # (or set in run-daemon.ps1 on Windows)
./install-mac.sh                       # or your platform's installer
# Daemon prints bearer token at startup:
#   "WiFi-bridge token: <copy this into wifi_config.h>"

# Firmware side
cp firmware/src/wifi_config.h.example firmware/src/wifi_config.h
# Edit wifi_config.h with SSID / password / host IP / token
pio run -e waveshare_amoled_216_wifi -t upload
```

Happy to split this into a smaller cleaner PR off main (or a stacked PR against #26) if that's easier to review — let me know what works.


alasdaircs:
Related: I built a variant that takes a different architectural position — the ESP32 calls the Anthropic API directly over Wi-Fi using a long-lived token, so there's no host daemon at all (not even an HTTP bridge). The BLE HID keyboard is kept for the physical buttons; the GATT data service is removed.

It also implements the captive-portal provisioning you flagged as a known follow-up: on first boot the device starts a softAP named "ClawdMeter", serves a setup form at 192.168.4.1, and stores credentials to NVS. Re-provisioning is triggered from a Wi-Fi screen on the device.

Fork is here if it's useful context: https://github.com/alasdaircs/clawdmeter-wifi

@HermannBjorgvin I'm happy to leave it as a standalone fork — just wanted to surface it given the overlap with this issue. If a PR that adds this as an opt-in build env would be welcome, say the word and I'll write one up properly.

PR #25, opened 2026-05-20
Add AMOLED-1.8 board + device-abstraction refactor for community ports

tobby168:
## Summary

Two related changes that together make the project welcoming to new
board ports:

1. **Add Waveshare ESP32-S3-Touch-AMOLED-1.8 support** (the existing
   commit you'd already see in my fork — SH8601 + FT3168 + XCA9554,
   368×448 portrait).
2. **Device-abstraction refactor** that follows it: a thin HAL in
   `firmware/src/hal/` + per-board folders under `firmware/src/boards/`
   instead of the build-flag-driven `#ifdef BOARD_*` sprawl the second
   port introduced. Shared code (`main.cpp`, `ui.cpp`, `splash.cpp`) is
   now `#ifdef`-free.

After this lands, adding a third board (or letting somebody else add
theirs) is `cp -r firmware/src/boards/template firmware/src/boards/<new>/`,
edit the TODOs, add a PlatformIO env. No shared file touches.

## What changed

**HAL (`firmware/src/hal/`)** — public interfaces shared code calls:
`display_hal`, `touch_hal`, `input_hal`, `power_hal`, `imu_hal`,
`board_caps`. Compile-time selection via PlatformIO's `build_src_filter`,
not virtual functions — no vtable cost.

**Per-board folders (`firmware/src/boards/`)**:
- `waveshare_amoled_216/` — CO5300 + CST9220 + AXP PKEY + QMI8658 rotation
- `waveshare_amoled_18/` — SH8601 + FT3168 + AXP + XCA9554 (PWR via EXIO4), no rotation
- `template/` — copy-this-to-start-a-port skeleton

Each board owns its `board.h` (pins, I2C addrs, `BOARD_HAS_*` flags),
`board_init.cpp` (Wire.begin + any IO expander release), `display.cpp`,
`touch.cpp`, `input.cpp`, `power.cpp`, `imu.cpp`, `caps.cpp`, plus any
board-private hardware drivers.

**Responsive UI**: `ui.cpp::compute_layout()` picks fonts and layout
from `board_caps()` at runtime — current breakpoints `H ≥ 460` → large,
else compact. `splash.cpp` computes `CELL = min(W, H) / 20` instead of
the two hardcoded values from the previous port.

**Capability flags**: `BoardCaps` (runtime struct) for things shared
code queries; `BOARD_HAS_*` (compile-time macros) for things the linker
can dead-strip per board. See [`docs/porting/capability-flags.md`](docs/porting/capability-flags.md).

[...]

PR #13, opened 2026-05-16
Feature/sensecap indicator d1l

carroarmato0:
Added the Seeed Studio SenseCap Indicator D1L model as a target.
Since this device only has one physical button at the top, that one is used to turn the screen on/off.
The rest is touch screen driven.
No internal battery -> hide the battery icon.

IMG20260516161533

PR #46, opened 2026-05-29
feat(boards): Add Seeed SenseCap Indicator D1L support

carroarmato0:
## Summary

This PR adds firmware support for the [Seeed Studio SenseCap Indicator D1L](https://wiki.seeedstudio.com/SenseCAP_Indicator_Get_Started/), following the HAL abstraction introduced in PR #25. It supersedes the earlier PR #13 which used `#ifdef TARGET_SENSECAP` throughout shared code.

## Hardware

| Component | Details |
|---|---|
| SoC | ESP32-S3, OPI PSRAM, 8 MB flash |
| Display | 480×480 ST7701S, RGB parallel interface |
| Touch | FT6x36 at I2C 0x48 (non-standard address), panel mounted 180° |
| IO expander | PCA9535 at I2C 0x20 — gates LCD SPI CS (P04) and touch RST (P07) |
| Button | GPIO 38, active-LOW, drives screen cycle / wake via `power_hal_pwr_pressed` |
| Battery / PMU | None |
| IMU | None |

## New files

All board-specific code lives in `firmware/src/boards/sensecap_indicator_d1l/`:

- `board.h` — pin definitions, I2C addresses, `BOARD_HAS_*` flags
- `io_expander.h` / `io_expander.cpp` — minimal PCA9535 raw-I2C driver (port 0 only, no external library)
- `board_init.cpp` — `Wire.begin` + `io_expander_init` (must run before `display_hal_init`)
- `caps.cpp` — `BoardCaps` instance
- `display.cpp` — `Arduino_RGB_Display` driver; ST7701S init sequence over SPI gated by PCA9535 CS; PWM backlight on GPIO 45
- `touch.cpp` — vendored minimal FT6x36 reader; 50 Hz poll (no INT pin); both axes flipped for 180° mount
- `power.cpp` — edge-detected GPIO 38 PWR button at 50 Hz; all battery stubs return -1/false
- `input.cpp` — stubs returning false (device has no microphone; PTT not applicable)
- `imu.cpp` — no-op stubs

## Shared-code changes (`firmware/src/ui.cpp`)

Two board-agnostic fixes proposed for upstream:

1. **Battery icon respects `board_caps().has_battery`** — `apply_battery_visibility()` was documented to hide the battery widget on boards without a battery, but the check was never implemented. Fixed.

2. **Swipe gesture navigation** — left/right swipes cycle screens; tap still toggles splash. Works on all boards. Also fixes a LVGL 9 quirk where a swipe fires a spurious `LV_EVENT_CLICKED` at lift-off (400 ms debounce guard in `global_click_cb`). `make_panel()` now explicitly sets `LV_OBJ_FLAG_GESTURE_BUBBLE` alongside `EVENT_BUBBLE`.

## Hardware notes

- **ST7701S colour register**: `0xCD = 0x08` (not `0x00` as in the Arduino_GFX type9 default — wrong value produces a permanent green tint).
- **No `INPUT_BTN_PRIMARY`**: the D1L has no BOOT/GPIO0 button available as a general input. GPIO 38 is repurposed as the PWR button via `power_hal_pwr_pressed`; voice-mode PTT is permanently false, which is correct since the device has no microphone.
- **`power_hal_is_vbus_in()` returns false**: the device has no USB VBUS detection. This intentionally allows the idle system to sleep the display after the timeout — desirable for a desk device that may run 24/7.
- **RGB parallel (non-QSPI) display**: the porting guide mentions QSPI AMOLED panels, but the HAL contract is interface-only and accommodates the RGB parallel driver without any HAL changes.

## Test plan

- [x] All three envs compile (`sensecap_indicator_d1l`, `waveshare_amoled_216`, `waveshare_amoled_18`)
- [x] Flashed and booted on hardware — splash animation visible, PCA9535 init OK, FT6x36 touch init OK
- [x] Button (GPIO 38) cycles screens and wakes from sleep
- [x] Touch swipe navigates screens; tap toggles splash
- [x] Battery icon hidden (no battery on this board)
- [x] No regressions on existing board envs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ firmware/.vscode/
 # Claude Code
 .claude/
 
+# Git worktrees
+.worktrees/
+
 # Node tooling deps (npm install --no-save)
 tools/node_modules/
 tools/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,10 @@ firmware/.vscode/
 tools/node_modules/
 tools/package-lock.json
 
-# Mac daemon venv
+# Python venvs
 daemon/.venv/
+firmware/.venv/
+firmware/.pio_home/
 
 # Python bytecode cache
 __pycache__/
--- a/firmware/platformio.ini
+++ b/firmware/platformio.ini
@@ -177,3 +177,50 @@ lib_deps =
     lvgl/lvgl@^9.2.0
     bblanchon/ArduinoJson@^7.0.0
     h2zero/NimBLE-Arduino@^2.1.1
+
+
+[env:sensecap_indicator_d1l]
+platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.38-1/platform-espressif32.zip
+board = esp32-s3-devkitc-1
+framework = arduino
+board_build.arduino.memory_type = qio_opi
+board_upload.flash_size = 16MB
+board_upload.maximum_size = 16777216
+board_build.partitions = default_16MB.csv
+upload_speed = 921600
+monitor_speed = 115200
+
+build_src_filter =
+    +<*>
+    -<boards/>
+    +<boards/sensecap_indicator_d1l/>
+
+build_flags =
+    -DBOARD_SENSECAP_INDICATOR_D1L
+    -DARDUINO_USB_CDC_ON_BOOT=1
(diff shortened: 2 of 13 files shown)


carroarmato0:
IMG20260529141426

PR #48, opened 2026-05-30
Add brightness adjustment screen (AMOLED-2.16)

JesusValeraDev:
## Summary

Adds a **Brightness** screen to the AMOLED-2.16, opened with the SECONDARY (GPIO18) button, with on-screen −/+ touch controls (5% per tap, 10% floor). Brightness is now a runtime value owned by the idle module, so it survives sleep/wake and screen rotation instead of snapping back to the compile-time default.

<table>
  <tr>
    <td>
      brightness-78
    </td>
    <td>
      brightness-40
    </td>
  </tr>
</table>

## Changes

- **idle.cpp / idle.h** — brightness becomes a runtime value (`idle_set_active_brightness` / `idle_get_active_brightness`); idle restores this on wake instead of the fixed constant.
- **ui.cpp / ui.h** — new `SCREEN_BRIGHTNESS` with a percentage readout, bar, and two −/+ touch buttons; the SECONDARY button toggles it open/closed.
- **main.cpp** — SECONDARY button now opens the brightness screen (was HID Shift+Tab).
- **display.cpp** — the rotation brightness ramp restores the active brightness rather than jumping back to a hardcoded full-bright value.
- **touch.cpp** — fix touch coordinates not following the software display rotation; per-quadrant transform (`3 − rotation`) measured on-device.

## Testing

- Builds clean for both `waveshare_amoled_216` and `waveshare_amoled_18`.
- −/+ buttons verified working on hardware in all four IMU rotation orientations.

## Notes

- Brightness is RAM-only — it resets to the default on a full power-cycle (NVS persistence not included in this PR).

### DIFF
--- a/firmware/src/boards/waveshare_amoled_216/display.cpp
+++ b/firmware/src/boards/waveshare_amoled_216/display.cpp
@@ -14,6 +14,10 @@ static uint16_t* rot_buf = nullptr;
 static Arduino_DataBus* bus = nullptr;
 static Arduino_CO5300*  gfx = nullptr;
 
+// Last brightness idle asked for; the rotation ramp restores this instead of
+// jumping to full bright, so a dimmed screen stays dimmed across a rotate.
+static uint8_t target_brightness = 200;
+
 void display_hal_init(void) {
     bus = new Arduino_ESP32QSPI(
         LCD_CS, LCD_SCLK, LCD_SDIO0, LCD_SDIO1, LCD_SDIO2, LCD_SDIO3);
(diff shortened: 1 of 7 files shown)


Chemaclass (approved):
Nice

HermannBjorgvin:
Thanks for the PR! I was working on the UX on a separate branch recently and it actually included brightness controls as well in #49 which also removes the Bluetooth settings screen. Could you perhaps review that one and tell me what you think?

I'm simplifying the UX to only have a single usage screen and the splash screen so this overlaps a bit with that but I'd love to get your review on it if you have the time :)

JesusValeraDev:
Great, I am flashing into my device right now.

We can close this issue then, and I will reply in PR #49 with my findings 😄

PR #51, opened 2026-05-31
Fix: macOS daemon can't find the device (HID-held peripheral hidden from scans)

soapiece:
## Problem

On macOS the Python daemon never finds the device — it logs `Device not found` forever even though "Claude Controller" is plainly visible in **System Settings → Bluetooth**.

Root cause is macOS-specific: the firmware advertises as a BLE **HID keyboard** (for the Space / Shift+Tab buttons), so macOS auto-connects to it as a keyboard. **CoreBluetooth excludes already-connected peripherals from scan results**, and bleak's connect-by-address path *also* scans internally (`find_device_by_address`), so a cached address can't help either. The scan-by-name approach works on Linux/BlueZ but can never see the device on macOS once the OS holds the HID link.

## Fix

On macOS (`sys.platform == "darwin"`), discover the target via CoreBluetooth's `retrieveConnectedPeripheralsWithServices_` — the documented API for reaching a system-connected peripheral — and connect to it directly (no scan). The custom data-service UUID is matched first (unambiguous); the generic HID service `0x1812` is only trusted on an exact `DEVICE_NAME` match so it can't grab an unrelated keyboard/mouse. A peripheral that fails to connect is skipped for one cycle so the scan fallback stays reachable.

The device's two BLE connection slots let the OS HID link and the daemon run **simultaneously** — the keyboard buttons keep working while the daemon streams usage.

**Linux/BlueZ path is unchanged** (still scans by name + caches the address); a redundant per-reconnect address save was removed so caching happens only on a fresh scan.

## Testing

Verified on macOS (Sequoia, Python 3.14, bleak 3.0.2) with a Waveshare AMOLED-2.16: both a foreground smoke test (`daemon/test_macos_connect.py`, added here) and the launchd background service connect and stream live data, and survive device reboot/reconnect — all while the device's HID keyboard buttons keep working. No regression on the Linux code path (logic unchanged).

### DIFF
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -157,6 +157,120 @@ async def scan_for_device() -> str | None:
     return None
 
 
+# --- macOS: recover a device the OS already holds as an HID keyboard --------
+#
+# The firmware advertises as a BLE HID keyboard so its buttons type into the
+# Mac. macOS auto-connects to that HID, and CoreBluetooth then EXCLUDES the
+# peripheral from BleakScanner.discover() results (already-connected devices
+# never appear in scans). bleak's connect-by-address path also scans
+# internally, so a cached address can't help either. The documented escape
+# hatch is retrieveConnectedPeripheralsWithServices_, which returns
+# peripherals the system is already connected to. We wrap the result in a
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
Thanks for the PR and the testing!

PR #52, opened 2026-05-31
Feature: display multiple Claude accounts with auto-rotation

soapiece:
## What

Show usage for **multiple Claude subscription accounts** on the device, not just one. The Usage screen auto-rotates through each account (every 5s) with a row of page dots, and the middle (PWR) button steps through them manually before moving on to the Bluetooth screen.

> **Note:** this is stacked on #51 (the macOS BLE fix). Until that merges, the diff here also includes its commit; it'll collapse to just the multi-account commits once #51 is merged.

## How it works

**Daemon** — accounts are declared in `~/.config/claude-usage-monitor/accounts.json` as a list of `{label, config_dir}`. No file = a single default account (fully backward compatible). Each account's token is resolved from its Claude Code profile:

- **macOS:** the per-profile Keychain item — `Claude Code-credentials` for the default profile, or `Claude Code-credentials-<hash>` where `<hash>` is the first 8 hex of `sha256(<absolute config_dir>)`. (This is how Claude Code itself namespaces `CLAUDE_CONFIG_DIR` profiles.)
- **Linux:** `<config_dir>/.credentials.json`.

All accounts are polled concurrently each cycle and sent as a single JSON array on the RX characteristic:

```json
{ "a": [ { "n": "main", "s": 45, "sr": 120, "w": 28, "wr": 7200, "st": "allowed", "ok": true } ] }
```

A failed account is still included as `{ "n": "...", "ok": false }` so the device shows it **offline** rather than dropping it. Writes use write-with-response so CoreBluetooth performs a long write for multi-account arrays up to the 512-byte device buffer.

**Firmware** — `parse_json()` reads the `{"a":[...]}` array into a fixed `accounts[]` (up to `MAX_ACCOUNTS = 8`, bounded by the BLE buffer), still accepting a legacy bare object. The usage screen renders one account at a time (label as the title), auto-rotates in `ui_tick_anim()`, and draws a page-dot indicator. Offline accounts show `--%` / `offline`.

## Backward compatibility

With a single account, behavior is **identical** to before: no page dots, title stays "Usage", no auto-rotation, the same Usage↔Bluetooth toggle, and the legacy single-object payload still parses (a missing `ok` defaults to `true`).

## Testing

Verified end-to-end on a Waveshare AMOLED-2.16 with two real accounts: the daemon polls both concurrently and the device auto-rotates between them with correct per-account numbers and page dots; manual button cycling works; a transient poll failure correctly rendered the account offline. Both the daemon and firmware diffs were reviewed by a second model (Codex) and its findings applied (first-payload rotate-timer reset; legacy `ok`-missing compatibility). Docs + `daemon/accounts.example.json` included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The Clawd animations come from [claudepix](https://claudepix.vercel.app), [@amaa
 
 ## Screens
 
-The device boots into the splash and stays there until you press the middle (PWR) button, which cycles between Usage and Bluetooth. Tap the screen anywhere (except the Reset zone on the Bluetooth screen) to flip back to the splash; tap again to dismiss it.
+The device boots into the splash and stays there until you press the middle (PWR) button, which cycles between Usage and Bluetooth. With more than one account configured (see [Multiple accounts](#multiple-accounts)), the Usage screen auto-rotates through each account and the middle button steps through them before moving on to Bluetooth. Tap the screen anywhere (except the Reset zone on the Bluetooth screen) to flip back to the splash; tap again to dismiss it.
 
 |              Splash               |              Usage              |                Bluetooth                |
 | :-------------------------------: | :-----------------------------: | :-------------------------------------: |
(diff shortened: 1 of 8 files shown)


HermannBjorgvin:
Getting an Enterprise account at work soon so I'll look into this sorry I haven't gotten around to your PR yet.

PR #49, opened 2026-05-30
Collapse UX to single Usage view + hold-to-pair + brightness

HermannBjorgvin:
🤖AI Generated:

## Summary

Collapses the device UX from three screens (Splash / Usage / Bluetooth) to two
(Splash / Usage). The Bluetooth screen's affordances are replaced by single-
view UX, layered on top of `main`'s responsive `Layout` + per-board HAL — no
architecture changes.

- **Hold-to-pair** (3 s hold then release) replaces the BT screen's "Reset
  Bluetooth" button. Clearing happens on *release* in a 3–6 s window, so
  holding to power the device off (8 s AXP shutdown) leaves the bond intact.
  Required extending the `power_hal` contract with long-press + release edges:
  216 via AXP PKEY `LONG`/`POSITIVE` IRQ; 1.8" via software hold-timing off
  the IO-expander poll (template board stubs the new functions).
- **Status line** at the bottom of Usage replaces the BT-screen status. Reuses
  the existing whimsical spinner: shows `Connected…` for 5 s after connect →
  whimsical messages while idle; `Disconnected…` when bonded;
  `Pairing…` when no bonds + advertising (via new `ble_has_bonds()`).
- **Pairing hint** (`To pair / hold the power button / for 3 seconds, then
  release`) fills the screen while disconnected so it isn't empty.
- **PWR short-press on Usage** now cycles screen brightness (4-step ramp
  {64,128,200,255}, default 200 = unchanged from `DISPLAY_DEFAULT_BRIGHTNESS`,
  persisted to NVS). New `brightness.{h,cpp}` integrates with `idle` via a new
  `idle_set_awake_brightness()` so idle's fade-to-awake target *is* the user's
  level — no fighting between brightness control and the idle fade system.
- **Splash creature** re-tinted from claudepix's lighter `#CD7F6A` to brand
  terracotta `#D97757` via a `convert_to_c.js` `TINT_OVERRIDE` (survives
  re-scrapes); `splash_animations.h` regenerated.
- **BLE device name** renamed from "Claude Controller" to **"Clawdmeter"**
  (firmware + both daemons + docs + install scripts).
- README: dropped the BT screen + screenshot from the Screens table, fixed the
  stale "Usage ↔ Bluetooth" cycling prose, documented the 3 s pairing gesture.

`backup/clawdmeter-ux-collapse-pre-rebase` is a local-only snapshot of the
pre-rebase work (the original branch before this port onto main's refactor).

## Test plan

[...]

PR #55, opened 2026-06-02
Add ESP32-S3 4848S040 board

ginkage:
Looks something like this:
image
The one thing with this board is that it doesn't have any buttons, no modification to the sources was needed, everything remains functional.


HermannBjorgvin:
Boards with more limitations like no button support should work a bit better after #49 is merged which collapses the UI into only the usage screen and splash screen.

Apologies to anyone who sees this that is also waiting on a PR review BTW

There's been a lot of them coming in and work has kept me a bit busy, did not expect the project to get so popular.

HermannBjorgvin:
Maybe add it to the board list in the readme and context files?

HermannBjorgvin:
The HAL abstraction should also allow you to hide the battery in case this is a non battery powered device, looks like that to me from the picture.

ginkage:
> Maybe add it to the board list in the readme and context files?

Done.

> The HAL abstraction should also allow you to hide the battery in case this is a non battery powered device, looks like that to me from the picture.

It's already used. Though the board does have battery connector, it doesn't come with a battery by default.

ginkage:
OK, sorry for some noise earlier today. This has now been rebased, reimagined, outstanding issues fixed, the whole thing properly solidified and documented. Still works, even better now. :)

PR #59, opened 2026-06-03
fix(windows-daemon): connect to renamed, bonded Clawdmeter

skie97:
## Summary

- **Rename** `DEVICE_NAME` from `"Claude Controller"` to `"Clawdmeter"` so the scan matches the device's current advertisement name (renamed in 13cc423).
- **Bonded-address fallback** (`acquire_target` / `discover_bonded_address`): once the device is paired as a bonded HID keyboard it stops advertising, so a plain scan loops forever on \"Device not found\". Recovers the address from the Windows PnP table and hands `BleakClient` a `BLEDevice` so WinRT connects directly via `from_bluetooth_address_*` without re-running a doomed scan.
- **Test seam**: repoints the `main()`-driving reconnect/tray tests at `acquire_target` so they no longer fall through to real bonded discovery and live BLE.
- **Docs**: documents the scan→bonded sequence and the `CLAWDMETER_BLE_ADDRESS` env-var override in `README-windows.md`.

## Test plan

- [x] Pair Clawdmeter with Windows as a bonded HID keyboard, restart the daemon — confirm it connects via the bonded-address path (log line: \"Not advertising; connecting to bonded address ...\")
- [x] Confirm `CLAWDMETER_BLE_ADDRESS` override bypasses the scan entirely

### DIFF
--- a/daemon/README-windows.md
+++ b/daemon/README-windows.md
@@ -42,7 +42,7 @@ before running the daemon:
 
 1. Put the device on its Bluetooth waiting screen (powered on, not yet connected).
 2. Open **Settings → Bluetooth & devices → Add device → Bluetooth**.
-3. Select **Claude Controller** and complete pairing.
+3. Select **Clawdmeter** and complete pairing.
 
 **Why this is required:**
 
@@ -104,8 +104,8 @@ python daemon\claude_usage_daemon_windows.py
 ```
 [HH:MM:SS] === Claude Usage Tracker Daemon (BLE, Windows) ===
 [HH:MM:SS] Poll interval: 60s
-[HH:MM:SS] Scanning for 'Claude Controller' (8.0s)...
-[HH:MM:SS] Found: XX:XX:XX:XX:XX:XX
+[HH:MM:SS] Scanning for 'Clawdmeter' (8.0s)...
+[HH:MM:SS] Not advertising; connecting to bonded address XX:XX:XX:XX:XX:XX
 [HH:MM:SS] Connecting to XX:XX:XX:XX:XX:XX...
 [HH:MM:SS] Connected
 [HH:MM:SS] Sending: {"s":42,"sr":180,"w":17,"wr":8820,"st":"active","ok":true}
@@ -114,6 +114,11 @@ python daemon\claude_usage_daemon_windows.py
 - **The device must be paired with Windows first** (see [Pair the device](#pair-the-device-one-time)).
   The daemon then connects over that existing link via `BleakScanner` + `BleakClient`; it does
   not pop its own pairing dialog.
+- **`Scanning…` → `Not advertising; connecting to bonded address …` is normal.** Once paired,
+  Windows keeps the device connected, so it stops advertising and an advertisement scan can't
+  see it. The daemon detects this and connects directly to the device's address (recovered from
+  the Windows PnP table). The 8-second scan that precedes the fallback happens once per session.
+  Set `CLAWDMETER_BLE_ADDRESS=AA:BB:CC:DD:EE:FF` to pin the address and skip PnP lookup.
 - After `Connected`, the daemon polls the Anthropic API immediately and sends the first
   payload within a few seconds of connect (warm token path). With a valid, non-expired token
   the device should leave its waiting screen and show session + weekly percentages within
@@ -135,7 +140,7 @@ Press **Ctrl+C** in the terminal. The daemon logs `Daemon stopping` and exits cl
 | Symptom | Likely cause | Fix |
 |---------|-------------|-----|
 | `Warning: running under Linux/WSL` | Running in WSL, not native Windows | Run from a native PowerShell or Command Prompt on Windows |
-| `Scanning for 'Claude Controller'… Device not found` | Clawdmeter is off, out of range, or showing a non-Bluetooth screen | Power on the device and ensure it is on the Bluetooth waiting screen |
+| `Scanning for 'Clawdmeter'… Device not found` | Clawdmeter is off, out of range, not yet paired, or showing a non-Bluetooth screen | Power on the device, pair it once (see [Pair the device](#pair-the-device-one-time)), and ensure it is in range |
 | `No token; skipping poll` | No credentials file found at any candidate path | Confirm `claude login` ran on this machine; check `%USERPROFILE%\.claude\.credentials.json` exists |
 | `API HTTP 401` | Token expired | Re-run `claude login` in a terminal to refresh the token, then restart the daemon |
 | `Connection failed` | WinRT BLE initialisation issue | Ensure Windows Bluetooth is on; try toggling Bluetooth off/on in Windows Settings |
--- a/daemon/claude_usage_daemon_windows.py
+++ b/daemon/claude_usage_daemon_windows.py
@@ -14,6 +14,7 @@ import logging.handlers
 import os
 import re
 import signal
+import subprocess
 import sys
 import threading
 import time
@@ -21,9 +22,10 @@ from pathlib import Path
 
 import httpx
 from bleak import BleakClient, BleakScanner
+from bleak.backends.device import BLEDevice
 from bleak.exc import BleakError
 
-DEVICE_NAME = "Claude Controller"
+DEVICE_NAME = "Clawdmeter"
 SERVICE_UUID = "4c41555a-4465-7669-6365-000000000001"
 RX_CHAR_UUID = "4c41555a-4465-7669-6365-000000000002"
 REQ_CHAR_UUID = "4c41555a-4465-7669-6365-000000000004"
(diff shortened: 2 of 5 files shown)


HermannBjorgvin:
Thanks for the Daemon fixes!

Could you explain the overage fix in a bit more detail and why you prefer the unified-5h header, what's the difference?

I'm on holiday probably won't get back to my computer until tomorrow so unable to test it more in depth at the moment.

HermannBjorgvin:
🤖 AI Generated:

Question on the overage-mode UX: once overage triggers, both `sr` and `wr` fall back to the same `unified-reset` value:

```python
"sr": reset_minutes(raw("5h-reset") or unified_reset),
"wr": reset_minutes(raw("7d-reset") or unified_reset),
```

So the device shows the 5h and 7d windows with **identical reset countdowns**, both reflecting the representative-claim (overage) reset rather than the real per-window boundaries — including the "when does my next session block free up" timer, which is arguably what a user in overage cares about most. Is that the intended experience, or should the per-window resets stay distinct? It hinges on whether the API actually *drops* `-5h-reset`/`-7d-reset` in overage or just drops `-utilization` — did your testing capture the raw header set?

Related, on the status field: `st` now prefers `unified-status` over `5h-status`, which changes the value sent in the **normal** (non-overage) path too, not only overage. Is that an intended behavior change, and what status string should the device expect to display in overage — `overage`, and does the firmware have a UX for that state?

skie97:
🤖 Co-written with Claude

Quick context on what this PR actually fixes, then your two questions.

**The bug:** the daemon reads usage from HTTP response headers, specifically `unified-5h-utilization` and `unified-7d-utilization`. When an account is in overage (included plan usage spent, now on pay-as-you-go), the API stops sending those headers entirely. The daemon found nothing, fell back to its `"0"` default, and sent `{"s":0,"w":0,"st":"unknown"}` — the empty payload this PR is about. It wasn't a parse bug or a bad token; the headers just aren't there for an overage account.

**The fix:** detect `overage-in-use: true` and send 100/100 + status `"overage"` instead of zeros (in overage, both windows are maxed by definition — that's why overage is active). Non-overage accounts are untouched; they still get the real 5h/7d numbers.

Here's the full unified header set captured live from the overage account — note there are no `-5h-*` or `-7d-*` headers at all:

```
anthropic-ratelimit-unified-status: allowed
anthropic-ratelimit-unified-reset: 1782864000
anthropic-ratelimit-unified-representative-claim: overage
anthropic-ratelimit-unified-fallback-percentage: 0.5
anthropic-ratelimit-unified-overage-in-use: true
anthropic-ratelimit-unified-overage-status: allowed
anthropic-ratelimit-unified-overage-utilization: 0.0
anthropic-ratelimit-unified-overage-reset: 1782864000
```

**1. Identical sr/wr countdowns**

In overage the API drops the per-window reset headers too (`-5h-reset`/`-7d-reset`), not just utilization — see the dump above. The code already prefers the per-window value when it exists (`raw("5h-reset") or unified_reset`), so distinct resets are preserved whenever the API sends them. The collapse only happens when there's nothing per-window to show.

You're right that the result is misleading though: `unified-reset` here equals `overage-reset` (~27 days out, a billing-horizon reset), so the 5h bar reads "Resets in 27d" — not when your session frees up. The API isn't giving us the real 5h boundary in overage, so we can't show it accurately. I'd rather emit `sr`/`wr` as `-1` (firmware renders `---`) when the per-window reset is genuinely absent, instead of a confident-but-wrong number. One caveat: this is from a single overage account, so I wouldn't hard-assume every overage response omits the per-window resets — I'll grab a second sample.

---

**2. st source change**

Yes, it changes the normal path too, and that was deliberate — `unified-status` is the authoritative overall state per Anthropic's own client logic ([claude-code#12829](https://github.com/anthropics/claude-code/issues/12829)), where `5h-status` only covers the 5h window. They match in the common case and diverge when the binding window is 7d or overage. The `or 5h-status` fallback keeps old behaviour for responses without `unified-status`.

But practically it doesn't matter today: the firmware never displays status. `st` is parsed into `UsageData.status` (`main.cpp:112`) and then nothing reads it — `ui_update()` only draws the two % bars and reset timers. So the `unified-status` switch and the `"overage"` string have zero on-screen effect; the only visible sign of overage is both bars at 100% red. If you'd rather keep `st` on `5h-status` to minimise the normal-path change, fine — it's a no-op either way until the firmware actually renders it.

HermannBjorgvin:
I think it makes sense to display info on the overage case, I would want to keep and eye on it myself if I had mine turned on.

In those cases I agree it makes sense to display 100% in the 5h limits but the overage reset being shown for the "Resets in %time%" text for the 5h/7d windows imo isn't a good ux, I think it would make more sense to have the Daemon cache it when the headers become unavailable.

Do the request headers give us the overage usage which we could display in a dollar figure? It could be a nice UX if the 5h bar fills 100% and turns red to show the extra usage in $'s vs. continuing to show "100%" which becomes less relevant if you're in the overage mode.

What do you think?

skie97:
Hmm, you've raised really good points. I'll be honest, I just received my waveshare 2.16 c6 yesterday in the mail, and just made enough changes to make it at least work and display. I have two accounts and one was overaged. Let me have another stab at it tonight after work, and propose something.

skie97:
Apologies, my accounts are not in overage and so I can't test this until the weekend. But generally, the plan is like this.
There is no dollar data available at all. The full header set looks like this:
```
  anthropic-ratelimit-unified-status:              allowed
  anthropic-ratelimit-unified-representative-claim: five_hour   ← which window is binding
  anthropic-ratelimit-unified-5h-utilization:      0.03   (3%)   anthropic-ratelimit-unified-5h-status: allowed
  anthropic-ratelimit-unified-5h-reset:            1780600200
  anthropic-ratelimit-unified-7d-utilization:      0.01   (1%)   anthropic-ratelimit-unified-7d-status: allowed
  anthropic-ratelimit-unified-7d-reset:            1780704000
  anthropic-ratelimit-unified-overage-utilization: 0.0          anthropic-ratelimit-unified-overage-status: allowed
  anthropic-ratelimit-unified-overage-reset:       1780589400
  anthropic-ratelimit-unified-fallback-percentage: 0.5
  anthropic-ratelimit-unified-reset:               1780600200
```
and so the plan is to change the ui to this when it hits overage.

```
    NORMAL                              OVERAGE (UI design)
  +------------------------+        +------------------------+
  |        Usage           |        |     Extra Usage        |  ← smaller (tiempos_34)
  |                        |        |                        |
  | 42%           Current  |        | 100%            Plan   |
  | [####------------]     |        | [################]     |
  | resets in 2h           |        | allowance spent        |
  |                        |        |                        |
  | 10%            Weekly  |        | 23%       Extra Usage  |  ← from overage-utilization
  | [##--------------]     |        | [###-------------]     |
  | resets in 5d           |        | resets in X            | (5h/7d whichever is more limiting)
  +------------------------+        +------------------------+
```
I'll add the commit to the PR once i can confirm it looks right in overage, and if you're ok with this change.

skie97:
Pushed the overage commit to this PR (`ca0ff9e`).

Daemon side (Windows, macOS, bash) now carries the real overage figure on the wire — `o` from `anthropic-ratelimit-unified-overage-utilization` and `or` from the representative-claim unified reset — instead of faking both windows to 100%. Outside overage both are absent → 0, so there is no phantom Extra Usage bar.

Firmware follows the mock I sketched above, with one change: rather than shrinking the title to the existing `tiempos_34`, I added a dedicated in-between **48px Tiempos** face and rendered "Extra Usage" in red. 34 felt too small against the 56px "Usage" and 56 overflows, so 48 splits the difference. I nudged it down 4px so its (shorter) cap-height lands on the same baseline. Plan bar is 100%/red/"Allowance spent"; the second bar is driven by the real overage %, with the adaptive reset formatter handling the window.

Verified the layout on my 2.16 c6 against a live-overage account. Let me know if you are happy with the red/48px treatment or want it dialed back.

claude_extra

HermannBjorgvin:
That's an awesome design! 🔥

I like it!

I am wondering if we can cache the last known sr value, that way you can easily see how long until your next window starts.

And I think it would be better to keep the "Current Session" label instead of changing it to "Plan"

Also thought tha we could use the yellow background for the "Extra Usage" label to highlight that the weekly limit indicator has changed to the extra usage indicator.

What do you think?

skie97:
After your last reply, I decided to work this out from the beginning and carefully, since I now have more time over the weekend. There were three accounts I wanted to test. A Pro/Max account, and Enterprise Account, and an account (for whatever reason) that uses an API-key. The API-key has no usage window concept at all and doesn't even store the creds in the .credentials.json, but instead in the .claude.json in the parent folder. You can see the table below for the details. For completeness, I append the full findings of the probes, for the two other accounts in and out of an overage scenario. In the end, there are only three possibilities.

| Header (`anthropic-ratelimit-unified-…`) | Pro/Max | Pro/Max overage | Enterprise |
|---|---|---|---|
| `…-5h-utilization` | `0.08` | `1.0` | `—` |
| `…-5h-status` | `allowed` | `rejected` | `—` |
| `…-5h-reset` | `1780735200` | `1780769400` | `—` |
| `…-5h-surpassed-threshold` | `—` | `1.0` | `—` |
| `…-7d-utilization` | `0.01` | `0.11` | `—` |
| `…-7d-status` | `allowed` | `allowed` | `—` |
| `…-7d-reset` | `1781308800` | `1781308800` | `—` |
| `…-representative-claim` | `five_hour` | `five_hour` | `overage` |
| `…-status` | `allowed` | `rejected` | `allowed` |
| `…-reset` | `1780735200` | `1780769400` | `1782864000` |
| `…-fallback-percentage` | `0.5` | `0.5` | `0.5` |
| `…-overage-in-use` | `—` | `true` | `true` |
| `…-overage-status` | `allowed` | `allowed` | `allowed` |
| `…-overage-utilization` | `0.0` | `0.0` | `0.54` |
| `…-overage-reset` | `1780722600` | `1782864000` | `1782864000` |
| `…-upgrade-paths` | `—` | `upgrade_plan,overage` | `—` |
| **Derived `st`** | `allowed` | `rejected` | `overage` |
| **Derived BLE payload** | `{"s":8,"sr":207,"w":1,"wr":9767,"o":0,"or":207,"st":"allowed","ok":true}` | `{"s":100,"sr":56,"w":11,"wr":9046,"o":0,"or":56,"st":"rejected","ok":true}` | `{"s":100,"sr":35681,"w":100,"wr":35681,"o":54,"or":35681,"st":"overage","ok":true}` |

Those three possibilities are exactly the three screen states, and the discriminator falls straight out of the table:

- `5h`/`7d` present → **Pro/Max**. `overage-in-use` absent → normal **Usage**; present → the windows are still real, so it stays the two-pill screen and just retitles to **Extra Usage**.
- `5h`/`7d` absent but `overage-in-use: true` → usage-based **Enterprise**: no windows at all, only `overage-utilization` against the monthly dollar-spend limit → its own **Enterprise** screen.
- neither (the API-key case) → no usage-window concept; out of scope for these screens for now (it'd just read 0%/unknown).

The new commit wires that up: the daemons emit two fields — `oiu` (overage-in-use) and `acct` (`pro-max`/`ent`, derived from window presence) — and the firmware switches screens on them. Pro/Max overage keeps its two real pills and only changes the title. For Enterprise there was really only one pill worth showing — the dollar-spend gauge (on the normal green→amber→red ramp) — so I dropped the second pill entirely and replaced it with a row of the animations from the splash screen. Both verified on real hardware (a 2.16" C6).

clawd-extra usage
clawd-enterprise

skie97:
Hi! would you prefer me to break this into 2 PRs? I can rework this to be Windows BLE connectivity, a simpler commit, and break the overage commits into another PR.

HermannBjorgvin:
> Hi! would you prefer me to break this into 2 PRs? I can rework this to be Windows BLE connectivity, a simpler commit, and break the overage commits into another PR.

Sorry must have missed this, yes I think that would be helpful to split it as this point in case we might need to revert either in the future for some reason.

skie97:
## Summary

- **Rename** `DEVICE_NAME` from `"Claude Controller"` to `"Clawdmeter"` so the scan matches the device's current advertisement name (renamed in 13cc423).
- **Bonded-address fallback** (`acquire_target` / `discover_bonded_address`): once the device is paired as a bonded HID keyboard it stops advertising, so a plain scan loops forever on \"Device not found\". Recovers the address from the Windows PnP table and hands `BleakClient` a `BLEDevice` so WinRT connects directly via `from_bluetooth_address_*` without re-running a doomed scan.
- **Test seam**: repoints the `main()`-driving reconnect/tray tests at `acquire_target` so they no longer fall through to real bonded discovery and live BLE.
- **Docs**: documents the scan→bonded sequence and the `CLAWDMETER_BLE_ADDRESS` env-var override in `README-windows.md`.

## Test plan

- [x] Pair Clawdmeter with Windows as a bonded HID keyboard, restart the daemon — confirm it connects via the bonded-address path (log line: \"Not advertising; connecting to bonded address ...\")
- [x] Confirm `CLAWDMETER_BLE_ADDRESS` override bypasses the scan entirely

HermannBjorgvin:
Merging this fix, thanks @skie97 for the patience on this.

PR #61, opened 2026-06-04
fix(splash): fill full screen on PSRAM-less C6

skie97:
## Problem

On the ESP32-**C6** board the splash animation renders as a small centered square with a black border instead of filling the 480×480 panel. The S3 boards are unaffected.

The splash canvas needs a full 480×480 RGB565 framebuffer (460 KB) to fill the panel. The C6 has **no PSRAM**, so the previous code capped the canvas at 200×200 in internal SRAM — hence the border. The S3 boards work because their framebuffer lives in PSRAM.

## Fix

Render a tiny **20×20** buffer (~800 bytes, fits internal SRAM easily) and let LVGL **nearest-neighbour upscale** it to fill the panel:

- Integer scale factors (480/20 = 24×) keep the pixel art crisp (`antialias = false`).
- The source-centre pivot keeps the scaled image centered on screen.
- The sizing decision is extracted into a pure, dependency-free `splash_compute_geometry()` (`splash_geometry.h`) with host-runnable unit tests.

The **PSRAM (S3) path is unchanged** — it renders full-size at scale 1.0×, so on-screen output is byte-for-byte identical. The old `MAX_CELL_NO_PSRAM` cap and `cell < 4` clamp are removed.

## Verification

- ✅ `waveshare_amoled_216` (S3) builds & links — confirms LVGL 9.2 `lv_image_*` API.
- ✅ `waveshare_amoled_216_c6` builds & links; RAM use dropped to 28.5% (the 78 KB canvas is now 800 bytes).
- ✅ Full-screen splash **confirmed on C6 hardware**.

## Notes

- Unit tests in `firmware/test/test_splash_geometry/` are host-runnable (`g++ -std=c++17 -I ../../src test_main.cpp`); wiring a PlatformIO `native` env so `pio test` picks them up is a worthwhile follow-up.

### DIFF
--- a/firmware/src/splash.cpp
+++ b/firmware/src/splash.cpp
@@ -1,5 +1,6 @@
 #include "splash.h"
 #include "splash_animations.h"
+#include "splash_geometry.h"
 #include "theme.h"
 #include "usage_rate.h"
 #include "hal/board_caps.h"
@@ -9,8 +10,10 @@
 
 // 20×20 grid. CELL sized so the canvas fits the smaller display dimension —
 // the canvas is square and centered, so on portrait or letterboxed panels
-// it leaves vertical margin rather than cropping.
-#define GRID         20
+// it leaves vertical margin rather than cropping. On PSRAM-less boards the
+// buffer is rendered tiny (cell == 1) and LVGL scales it up to fill the panel;
+// the geometry decision lives in splash_compute_geometry() (splash_geometry.h).
+#define GRID         SPLASH_GRID
 static int  cell      = 24;        // recomputed in splash_init()
 static int  canvas_w  = GRID * 24;
 static int  canvas_h  = GRID * 24;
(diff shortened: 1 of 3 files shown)


skie97:
Follow-up: this fills the screen, but after running it on the C6 for a while the splash got annoying to look at — the animation was laggy, frames visibly hitching, and worse during the busier (dance) animations.

Dug into the cause: the tiny-canvas + `lv_image_set_scale` upscale makes LVGL software-transform the *whole* 480×480 on every redraw. Profiled it on hardware at **~0.76 µs/output-px → 100–220 ms per frame** on the single-core C6. (I also tried a dirty-rect via `lv_obj_invalidate_area`, but LVGL doesn't cleanly clip a transformed image's redraw — it kept re-running the full transform and smeared.)

Pushed `472676d` to fix it. On the no-PSRAM path it drops the LVGL transform entirely: upscale the 20×20 cells ourselves (nearest-neighbour replication into a ~23 KB strip buffer) and push only the **changed** cells straight to the panel via the display HAL. The splash is now bound only by the QSPI flush (~0.20 µs/px):

| frame | before | after |
|---|---|---|
| full switch | 143 ms | 30 ms |
| high-motion (dance) | 126 ms | 26 ms |
| blink | 2.4 ms | 0.3 ms |

~5× faster, smooth, and the dirty-rect is now exact so there's no smearing. The S3 (PSRAM) boards keep the original LVGL canvas path unchanged.

PR #63, opened 2026-06-05
daemon: pure free-ride on Claude Code's token (read-only, never refresh)

nathanjohnpayne:
Keeps the desk monitor honest when the Claude OAuth token expires, without freezing on stale numbers — and **without the daemon ever touching the OAuth refresh endpoint**.

**Approach: pure free-ride.** The daemon is a read-only consumer of Claude Code's credential. It reads whatever access token Claude Code currently holds (Keychain `Claude Code-credentials` on macOS, `~/.claude/.credentials.json` on Windows/Linux), polls, and **never refreshes the token itself**. On a 401 it emits `{"ok": false}` and the device shows its idle "No data" screen until **Claude Code** — the token's legitimate owner — re-seeds it (i.e. when you next use the CLI, or run `claude login`).

**Why no self-refresh** (evidence in-thread): two clients sharing one rotating refresh token is a documented anti-pattern — concurrent rotation trips reuse-detection / exhausts a shared rate-limit bucket, and only a fresh login (a different OAuth grant) resets it ([refresh-token rotation BCP](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation), [CodexBar #575](https://github.com/steipete/CodexBar/issues/575), [opencode #18329](https://github.com/anomalyco/opencode/issues/18329)). We saw exactly that: the daemon's refresh persistently `429`'d while Claude Code's own refresh succeeded and re-seeded the credential — which the daemon then free-rode on to recover. So the daemon stops competing entirely.

Includes:
- A `{"ok": false}` "no-data" beat so the device shows idle "No data" promptly when the token's dead (pairs with #65) instead of holding stale numbers
- macOS + Windows parity; **removes ~550 lines** of refresh / backoff / write-back machinery
- `test_freeride.py`: read-path coverage, a regression guard that neither daemon can refresh, and a loop test that a dead token yields a `{"ok": false}` beat

**Behaviour:** while you use Claude Code it keeps the token fresh and the daemon free-rides; when nothing has refreshed it (laptop closed, CLI unused), the device honestly shows "No data" until you next use Claude Code / re-login — a better outcome than a self-inflicted rate-limit.


HermannBjorgvin:
We had some discussions on this going in #32 and #39

I really am keen to avoid taking such an invasive action as refreshing a token on behalf of users since it interacts directly with Claude Code's credential store both without the user being aware and potentially breaking it in the future even though the chances of that are slim.

I believe the right move is to go into an idle state if the token doesn't get refreshed since if the token is not refreshed by the user then we can at least say for certain the user isn't using Claude Code.

WDYT?

nathanjohnpayne:
Pushed a small follow-on commit (`ab94b7a`) per the heartbeat discussion on #64: the daemon now sends a `{"ok": false}` "no-data" beat when it genuinely can't get data — token dead + refresh can't recover, or no token — instead of going silent. The device then shows its idle screen promptly rather than holding stale numbers for the full freshness window. Transient network / rate-limit misses stay silent, so a blip doesn't flicker it to idle. (macOS + Windows.)

Pairs with #65 — a tiny firmware change that interprets the beat. **#65 should land first** so deployed firmware understands `ok:false` (without it, firmware would read `{"ok":false}` as a usage payload and show 0%). Verified end-to-end: beat → idle within a second.

nathanjohnpayne:
Refined the refresh strategy (commit `956516c`): dropped **proactive** refresh in favor of **free-riding** on Claude Code's token.

Proactive refresh turned out to compete with the Claude app's own refreshes for the token endpoint and trip its rate limit (429) — which then blocked recovery and forced a manual `claude login` (the exact thing this PR was meant to avoid). Now the daemon just uses whatever token Claude Code currently holds and only refreshes reactively on a 401. While you use Claude, the app keeps the token fresh and the daemon never touches the refresh endpoint.

Updated the title/description to match. Verified live: token recovered and steady data flow resumed.

nathanjohnpayne:
## Update: switched the refresh strategy to "free-ride" — validated on macOS over ~16h

**TL;DR:** proactive token refresh turned out to be the wrong call — it competed with Claude Code / the Claude app for the same OAuth endpoint and tripped its rate limit (`429`), which then *blocked* recovery and forced a manual `claude login`. The fix is to stop competing: free-ride on Claude Code's token and only refresh reactively. Proven on my Mac (details below); **still needs a Windows tester.**

### What was tried
1. **Proactive + reactive refresh** — renew ~2 min before expiry every cycle, plus refresh-and-retry on a 401.
2. **Cooldown + exponential backoff + give-up notice** — added when a failing refresh hammered the endpoint (300 s floor, doubling to a 30 min cap, a "re-login" notice after 5 consecutive failures).
3. Root-caused the actual problem: the **proactive** refresh was the *source* of the 429. Claude Code (and the app) already refresh the shared credential; the daemon refreshing *on top of that* pushed the token endpoint over its rate limit — and the cooldown/backoff were only treating the symptom.

### How recovery worked before this change
The cooldown/backoff kept the daemon from *worsening* the 429, but they couldn't *clear* it — they just spaced out retries that kept failing (I watched one throttle persist ~5 hours). With no way to wait it out, recovery was **manual every time**: run `claude login` to re-seed a fresh credential into the Keychain (the daemon reads the token fresh each poll, so it picks the new one up within ~60 s; `launchctl kickstart -k "gui/$(id -u)/com.user.claude-usage-daemon"` forces an immediate re-read).

So the screen would sit on stale numbers — and, once the idle screen existed, "No data" — for an unbounded stretch until I noticed, re-authenticated by hand, and kicked the daemon. Removing that manual `claude login` loop for the normal (laptop-open) case is the whole point of this change.

### What worked: free-ride (reactive-only)
Drop proactive refresh entirely. The daemon now uses whatever access token Claude Code currently holds (Keychain item `Claude Code-credentials` on macOS) and refreshes **only reactively, on a 401**, as a last resort. While you use Claude Code, the app keeps that token fresh and the daemon never touches the refresh endpoint → no self-inflicted 429, no forced login.

**Tradeoff:** if nothing refreshes the token — laptop closed, or you only use the *desktop app* (which keeps a separate credential the daemon doesn't read) — the device shows an honest "No data" until you next use Claude Code. That's a better failure mode than the old self-inflicted rate-limit.

### Proof (macOS)
Environment:
- macOS 26.5 (build 25F71), Apple M1 Pro
- Python 3.14.5, httpx 0.28.1, bleak (CoreBluetooth backend)
- Device: Waveshare ESP32-S3 AMOLED-2.16 (CO5300, 480×480)

One continuous **~16.6 h** daemon session (15:37 → 08:16), **0 proactive refreshes**:

| Window | Mac state | Device |
|---|---|---|
| 15:37 → ~23:00 | open, using Claude Code | 🟢 live — a fresh reading **every minute, continuously, past the ~8 h token lifetime** |
| ~00:00 → 07:50 | **lid closed / asleep** (on AC; Maintenance-Sleep + DarkWake cycles) | 😴 honest "No data" — token went invalid (wall-to-wall `API HTTP 401: Invalid authentication credentials`), nothing running to refresh it |
| 07:51 (lid opened) → live | reopened; Claude Code resumes | 🟢 self-healed to live — **no `claude login`** |

The overnight gap lines up exactly with the macOS sleep log (lid wake `UserActivity` at `2026-06-06 07:51:37`). The daemon kept polling during the Mac's dark-wakes and correctly emitted `{"ok":false}` each time — so the screen was never lying, just honest about a sleeping host.

### ⚠️ Needs Windows testing
The Windows daemon got the same change (parity: no proactive refresh, reactive-only, same `{"ok":false}` beat) and the unit tests pass — but I can only validate the macOS / CoreBluetooth path on my hardware. **Someone on Windows should confirm** the WinRT/BLE daemon free-rides on Claude Code's Windows credential and behaves the same way.

### Relationship to the firmware change (#65)
This PR's `{"ok":false}` "no-data" beat is consumed by the firmware change in **#65** — that's what flips the device to the idle "No data" screen *immediately* instead of holding stale numbers for the freshness window. They're complementary:
- **Free-ride (this PR)** keeps the token alive during normal use, so the beat rarely needs to fire.
- **The beat + #65** make the device honest the instant the token genuinely can't be fetched (e.g., the closed-lid overnight window above).

**Merge order:** #65 should land **before** this PR, so deployed firmware understands `ok:false` — without it, firmware would read the beat as a usage payload and render 0%.

nathanjohnpayne:
Final pivot (and I think the right one): **pure free-ride** — the daemon now never touches the OAuth refresh endpoint at all.

Earlier this PR did reactive refresh (on a 401). But validated on hardware + cross-checked against the literature, that was still wrong: the daemon's refresh always `429`'d (the endpoint rate-limits a piggybacking second client), while Claude Code's *own* refresh succeeded and re-seeded the credential — which the daemon then free-rode on to recover. So the daemon's refresh was pure dead weight that only kept its own throttle warm.

Evidence that two clients sharing one rotating refresh token is the anti-pattern: refresh-token rotation reuse-detection ([Auth0](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation)), and a near-identical menu-bar tool that hit the exact "poll/refresh → 429-bucket exhaustion → manual re-login" cycle ([CodexBar #575](https://github.com/steipete/CodexBar/issues/575)).

So the daemon is now a pure read-only consumer: read Claude Code's token, poll, and on a 401 emit `{"ok": false}` (idle "No data") and wait for Claude Code to re-seed it. Removed ~550 lines of refresh/backoff machinery; added `test_freeride.py` (read-path + a regression guard that neither daemon can refresh + the no-data-beat loop test). macOS validated live; **still needs a Windows tester.**

nathanjohnpayne:
## Why this went all the way to "never refresh": the failure and the research

### The failure I hit
Even after switching to **reactive-only** refresh (refresh only on a 401, never proactively), the device still got stuck on "No data" and couldn't recover on its own. Captured live on macOS:

- The access token was minted at **07:52** and expired a clean **~8h later at 15:52** (last good reading 16:00, then dead).
- Nothing refreshed it at that moment, so the daemon hit 401s and tried its reactive refresh — which came back **`HTTP 429` "Rate limited"** every time. Failures accumulated until it sat in a 30-minute backoff, **unable to recover.**
- At **16:26 the credential was rewritten by Claude Code itself** (the CLI's own refresh, triggered by normal use) and the daemon's *next poll* free-rode straight back to live — **no daemon refresh involved.**
- One minute later, at **16:27, I forced the daemon's refresh by hand → `429` again.**

That last pair is the whole story: **the daemon's refresh is persistently rate-limited, while Claude Code's own refresh succeeds.** Same endpoint, same token, ~same minute. So the daemon's refresh was *dead weight* — it never worked, and every attempt only kept its own throttle warm. The thing that actually recovered the device was free-riding on the CLI's refresh.

### The research (why a second refresher can't win)
- **Refresh-token rotation + reuse detection.** Refresh tokens are single-use and rotated; presenting an already-rotated one trips reuse detection, which revokes the token family / exhausts a rate-limit bucket. **Two clients sharing one refresh token and both rotating it is the documented anti-pattern** — the correct design is "one owner refreshes, everyone else reads." ([Auth0 refresh-token rotation](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation), [token rotation overview](https://www.token.security/glossary/token-rotation))
- **A near-identical tool hit the identical wall.** CodexBar (a macOS menu-bar Claude usage indicator) documents the exact chain: it fell back to **polling/refreshing the OAuth endpoint → "token exhaustion through retry spam" → the rate limiter blocks the exhausted token bucket with `429`**, and the documented fix is `claude logout && claude login` to *reset the bucket*. That's our failure verbatim, including the recovery. ([CodexBar #575](https://github.com/steipete/CodexBar/issues/575))
- **The endpoint is just aggressively limited.** On OpenCode, the `console.anthropic.com/v1/oauth/token` `429` **persisted even after changing the User-Agent and reproduced across multiple accounts** — a genuinely tight limit, so the fewer non-owner calls hit it, the better. ([opencode #18329](https://github.com/anomalyco/opencode/issues/18329))
- **Even the tool that self-refreshes hedges.** [opencode-claude-auth](https://github.com/griffinmartin/opencode-claude-auth) proactively refreshes + writes back — but **"falls back to the Claude CLI if the direct refresh fails."** Nobody fully trusts a piggybacking client's own refresh.

### The fix
Stop competing. The daemon is now a **pure read-only consumer** of Claude Code's credential: read the stored access token, poll, and on a 401 emit `{"ok": false}` (the device shows an honest idle "No data") and wait for **Claude Code — the token's owner — to re-seed it** (next CLI use, or `claude login`). The daemon never touches `/oauth/token`, so it can't trip the rotation race or feed the 429 bucket. ~550 lines of refresh/backoff machinery removed; a regression guard now asserts neither daemon can ever refresh.

Validated on macOS; **still needs a Windows tester.**

nathanjohnpayne:
Let's see how long this fix holds. It's complex, but I really can't even with manual refreshes. 😄

HermannBjorgvin:
> Let's see how long this fix holds. It's complex, but I really can't even with manual refreshes. 😄 

Thanks for the help Nathan, I'm on holiday until the 10th, just been sneaking code reviews on my phone when I get the chance, ping me if you need any more input on this or more testing, @kvenanzi  opened a draft pull request on this topic as well hoping he'll be able to help us test on the windows side if he's able

HermannBjorgvin:
I'm back in action now @nathanjohnpayne sorry you caught me almost as soon as I started my vacation so I was not very active this week :)

I will test this tomorrow on my end and hopefully we can merge this, thanks for working on this hopefully I wasn't too hard to handle in the review.

BTW a tip on Claude Code I noticed that I hope you don't mind me mentioning, I have this happen to me as well that sometimes it will update tickets titles and descriptions and pull request descriptions after you do a pivot like you did it will reword it based on the current session context "(read-only, never refresh)" which only makes sense if you were there at the time the pivot happened but not if you come to it later.

It's a good instruction to add to have titles and descriptions reflective of the work performed and not the journey the chat session took.

nathanjohnpayne:
Now I have been on vacation, home for a few days, and back to traveling. Thanks for the tip! Claude does do that, and it makes it hard to follow along.

## PR and issue titles/descriptions: describe the work, not the session

Titles and descriptions—for both pull requests and issues—must describe the
final state of the change (what it does and why) for a reader who arrives with
no knowledge of the session that produced it.

- Do not narrate the session's path: no pivots, abandoned approaches,
  "originally did X, then switched to Y," or commentary on how the plan evolved.
- When a pivot changes *what the work is*, update the title/description to
  reflect the new end state—not the fact that a pivot happened.
- Once a title or description already describes the work accurately, treat it as
  read-only. Do not reword or "refresh" it to fold in later session context.

cosjef:
@nathanjohnpayne Have we considered these alternative approaches:

1) Parse Claude Code's local JSONL transcripts instead of calling the API at all. Claude Code writes per-session usage data (tokens, model, timestamps) to ~/.claude/projects/**/*.jsonl on disk. This is exactly how the popular ccusage tool computes usage stats: zero network calls, zero tokens, zero auth to expire. The daemon would just tail files. The limitation is that it reconstructs usage from local records rather than reading the account's official quota/limits endpoint, so if the device specifically displays "% of 5-hour window remaining" per Anthropic's server-side counters, local parsing approximates rather than mirrors it. But it eliminates the entire class of auth problems, and for a desk gadget, an approximation that never goes dark may beat exact numbers that die at 3am.

2) A pragmatic middle ground within the current design might be: on a 401, instead of only waiting, the daemon could invoke the CLI itself, e.g. shelling out to a trivial claude command, so the token owner performs its own refresh through its own code path. That keeps "one owner refreshes" intact while removing the wait-for-human dependency. It's slightly gross (spawning a CLI to side-effect a credential), and it burns a tiny bit of usage, but it directly fixes the overnight/idle gap. One project cited in the thread already hedges this way: opencode-claude-auth falls back to the Claude CLI if its direct refresh fails.

Thoughts?

cosjef:
@nathanjohnpayne Any thoughts here?

nathanjohnpayne:
Fully agree — and that's exactly where this landed. There's **no refresh path left in either daemon**: the daemon is now a pure read-only consumer of whatever token Claude Code currently holds, and it never touches the OAuth refresh endpoint (there's even a regression test asserting neither daemon can). When the token isn't fresh — laptop closed, CLI unused — the device shows an honest **"No data"** idle screen until you next use Claude Code, rather than refreshing on your behalf or holding stale numbers. So it does precisely what you argued for here.

Validated on macOS over a continuous ~16.6 h run (0 proactive refreshes; overnight sleep → honest "No data" → self-healed to live on lid-open, no `claude login`). Just addressed the Copilot review nits too (the no-data beats now route through the zombie-link / write-failure path). Rebased on latest `main` and green — let me know if there's anything else you'd like before it goes in.

HermannBjorgvin:
Guys it's so hard to read AI essays these days I feel like the newest models have also degraded in their ability to explain things in a human way

Can a human write a tweet length description of what this does and why it should be merged?

HermannBjorgvin:
Had my agent recap me on this, going to just merge it since it's human verified.

PR #68, opened 2026-06-07
fix(windows-daemon): auto-refresh expired Claude OAuth token

kvenanzi:
## Problem

The Windows daemon reads only the short-lived `accessToken` from `.credentials.json` and never refreshes it. Because nothing renews the token while the PC is asleep or powered off, the first poll after wake gets a 401 and the tray repeatedly tells the user to run `claude login` — often several times a day.

## Fix

Implement the OAuth refresh-token flow Claude Code performs internally:

- **`get_valid_token()`** — proactively refreshes when the access token is expired or within ~5 min of `expiresAt`, before polling. Re-reads the credentials file immediately before refreshing so that if Claude Code refreshed concurrently we use its fresh token instead of spending our (possibly already-rotated, single-use) refresh token.
- **`_refresh_oauth_token()`** — POSTs the refresh token to the OAuth endpoint (`platform.claude.com`, with `console.anthropic.com` fallback) using the public Claude Code `client_id`; gates genuine auth failure vs transient purely on HTTP status.
- **Atomic write-back** to `.credentials.json` (same-dir `mkstemp` + `fsync` + `os.replace`) preserving every other key, so native Claude Code stays logged in and the rotated refresh token survives a daemon restart.
- **Reactive path** — one forced refresh + retry on an unexpected poll 401. The "run claude login" toast now fires only when the refresh token itself is genuinely revoked.

### Hardening
- A genuine refresh `AuthError` is caught in the poll loop (it previously could crash the daemon thread).
- An empty `access_token` in a 200 response is treated as transient and not written back (avoids burning the single-use refresh token in a loop).

## Scope & testing

Scope is the Windows daemon only (`daemon/claude_usage_daemon_windows.py` + tests). 124 daemon tests pass (31 new token tests, fully mocked — no network).

### DIFF
--- a/daemon/claude_usage_daemon_windows.py
+++ b/daemon/claude_usage_daemon_windows.py
@@ -15,6 +15,7 @@ import os
 import re
 import signal
 import sys
+import tempfile
 import threading
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
Check out #63 we're discussing possible solutions there.

If you could help test and review with us there that would be much appreciated.

Just a quick note the token refresh endpoint is rate limited and I have a general stance that we shouldn't ever refresh any token ourselves due to the possible unforseen problems it could cause in the future.

PR #58, opened 2026-06-03
amoled-18: auto-detect CO5300/CST816 panel revision

alphonsekoh:
## Problem

Waveshare ships two hardware revisions under the name **ESP32-S3-Touch-AMOLED-1.8**:

| | Original (what the firmware assumed) | Later revision |
|---|---|---|
| Display | SH8601 | **CO5300** |
| Touch | FT3168 @ I2C `0x38` | **CST816** @ I2C `0x15` |

On the later board the current firmware leaves the **panel blank** (SH8601 init/brightness commands are ignored by the CO5300) and spams `ESP_ERR_INVALID_STATE` on every touch (nothing answers at `0x38`).

Diagnosed on real hardware: an I2C scan found `0x15` (CST816, chip ID `0xB7`) with no device at `0x38`, and swapping in the `Arduino_CO5300` driver lit the panel. The LVGL framebuffer was rendering correctly the whole time.

## Fix

Detect the revision at boot by probing the touch address, then pick the matching drivers — **one binary drives both boards**:

- **`board_rev.{h}` / `board_init.cpp`** — probe `0x15` (CST816) vs `0x38` (FT3168); the touch chip is a reliable proxy for the panel module.
- **`display.cpp`** — instantiate `Arduino_CO5300` or `Arduino_SH8601` behind a shared `Arduino_OLED*`. The CO5300's 368-wide active area starts at GRAM column 16, so it gets a 16px column offset to center the image (verified centered on hardware).
- **`touch.cpp`** — both controllers share the FocalTech-style data layout at regs `0x02..0x06`, so a single reader serves both; only the I2C address changes by revision.
- **`board.h`** — add `CST816_ADDR 0x15`.
- **`CLAUDE.md`** — document the dual revision.

The original FT3168/SH8601 path is unchanged, so existing boards are unaffected.

## Testing

Flashed and verified on a CO5300/CST816 unit:

```
Board revision: CO5300 + CST816
Touch CST816 ID=0xB7 (addr 0x15)
Touch attached on INT pin
```

- Display: splash + UI render and are centered (16px offset).
- Touch: responds to taps/swipes; the previous `ESP_ERR_INVALID_STATE` spam is gone.
- Builds clean: `pio run -e waveshare_amoled_18`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #69, opened 2026-06-08
fix(c6): drive AMOLED-2.16 C6 panel as CO5300, side-USB orientation

skie97:
## Summary

`platformio.ini` (and the board files) labelled the Waveshare ESP32-C6-Touch-AMOLED-2.16 as using an **SH8601** panel, when it actually uses a **CO5300** — the same controller as the S3 AMOLED-2.16 — per the [official spec](https://docs.waveshare.net/ESP32-C6-Touch-AMOLED-2.16). It still rendered because `display.cpp` was patched with CO5300-style manufacturer registers (a `0xFE` page-switch writing the `0x19`/`0x1C` driving-voltage registers) grafted onto the `Arduino_SH8601` class, and both controllers speak standard MIPI DCS for pixel writes — so the panel lit up despite being driven through the wrong driver.

- **CO5300 display driver** (`9375bd6`): rebased onto `Arduino_CO5300` — matching the S3 AMOLED-2.16 sibling — keeping only the page-0x20 panel-driving registers (`0x19`/`0x1C`) that the class init omits; everything else (`0xC4`/`0x36`/`0x53`/`0x51`/`0x63`/`0x29`) is now covered by `Arduino_CO5300::begin()`.
- **Side-USB orientation**: adopts the CO5300 rotation-0 default (`MADCTL 0x00`) instead of the old force-written `0x30` transpose, orienting the board with the USB port on the side rather than the bottom — the preferred desk orientation.
- **Touch realign**: the touch axis mapping was calibrated to the old transposed raster, so it's re-derived and tap-tested on hardware to match the new orientation — raw CST9217 coordinates map straight through (`swap=false`, `mirror=(false, false)`).

The change is confined to the `waveshare_amoled_216_c6` board folder (plus the corrected `platformio.ini` comment) — no shared code, no other board affected.

## Test plan

- [x] `waveshare_amoled_216_c6` builds & links (RAM 28.5%, flash 21.6%)
- [x] Panel lights up and renders on **C6 hardware** — confirms the trimmed CO5300 init is sufficient
- [x] Orientation confirmed on hardware (USB port on the side)
- [x] Touch alignment tap-tested on hardware — dot tracks finger on both axes

### DIFF
--- a/firmware/platformio.ini
+++ b/firmware/platformio.ini
@@ -170,7 +170,7 @@ build_flags =
     -DLV_USE_SNAPSHOT=0
 
 lib_deps =
-    ; 1.6.4+ required for Arduino_SH8601 (this board uses SH8601, not CO5300)
+    ; this board uses the CO5300 (same controller as the S3 2.16); Arduino_CO5300
     moononournation/GFX Library for Arduino@^1.6.4
     lewisxhe/SensorLib@^0.2.6
     lewisxhe/XPowersLib@^0.2.7
--- a/firmware/src/boards/waveshare_amoled_216_c6/board.h
+++ b/firmware/src/boards/waveshare_amoled_216_c6/board.h
@@ -3,9 +3,9 @@
 // Waveshare ESP32-C6-Touch-AMOLED-2.16.
 //
 // Despite the matching "2.16" model number, this board is hardware-
-// distinct from the S3 AMOLED-2.16: it uses an SH8601 panel (same driver
-// family as the AMOLED-1.8), a CST9217 touch controller, and the C6 SoC's
-// own GPIO map. There is no PSRAM. AXP2101 PMU and QMI8658 IMU carry over.
+// distinct from the S3 AMOLED-2.16: it shares the CO5300 panel controller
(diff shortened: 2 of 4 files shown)


HermannBjorgvin:
Could this be a similar situation to #58 with different hardware revisions?

skie97:
Thanks for pointing me to that PR. I thought it was fascinating! My quick search revealed the following.

With respect to #58, the [english](https://docs.waveshare.com/ESP32-S3-Touch-AMOLED-1.8) docs for the S3-1.8 still point to the SH8601, but looking at the [chinese](https://docs.waveshare.net/ESP32-S3-Touch-AMOLED-1.8) docs, they mention a V2 for the device as of 30 May 2026 (8 days ago!) and it uses the CO5300 used instead. So #58 is correct. So the device in question was brand new.

For the C6-2.16 however, both the chinese and the english docs only list it as having one version, and it uses the CO5300. Thus, it seems to be a different scenario from #58.

HermannBjorgvin:
Thanks for investigating it. We'll go with your implementation then since it's the only documented one.

Thanks!

PR #72, opened 2026-06-10
feat: optional clock on the Usage screen (off by default)

devalnor:
Addresses #54 — re-opening the clock idea with a working implementation and screenshots, as you suggested.

## What

Optionally show the current time in place of the **Usage** title. It's **off by default**, so existing setups are completely unchanged. One config knob on the daemon turns it on.

| 24-hour | 12-hour | off (default) |
| --- | --- | --- |
| 24h | 12h | off |

## How it works

The firmware has no RTC, so the daemon is the time source. When enabled, the daemon adds two optional fields to the existing BLE JSON payload:

- `t` — local wall-clock epoch (UTC epoch shifted by the host's tz offset, so the firmware's `gmtime()` renders local time)
- `tf` — `12` or `24`, the hour format

The firmware caches `(epoch, lv_tick)` on each payload and **advances the clock locally** in `ui_tick_anim()`, so it ticks every minute between the 60s polls instead of only updating on each payload. The title label is only rewritten when the minute changes.

## Config

A single option, in a small `key = value` file the daemon re-reads every poll (no restart needed):

- macOS / Linux: `~/.config/claude-usage-monitor/config`
- Windows: `%LOCALAPPDATA%\Clawdmeter\config`

```
clock = off | auto | 12 | 24
```

- `off` (default) → no clock, the screen keeps showing "Usage"
- `auto` → show the clock, auto-detect 12h/24h from the host (macOS `defaults`, Linux `locale`, Windows registry)
- `12` / `24` → force the format

See `daemon/config.example`.

## Backward compatibility

Fully additive. Old daemons (no `t`) → firmware shows "Usage". Old firmware → ignores `tf`. No new dependencies, no `#ifdef BOARD_*` in shared code.

## Testing

Verified end-to-end on a real Waveshare ESP32-S3-Touch-AMOLED-2.16 for all three states (`12` → "6:49 PM", `24` → "18:51", `off` → "Usage" reverts live within one poll). Daemon config parsing unit-tested for bash + Python (off/auto/12/24, comments, CRLF, missing file). Both firmware envs (`waveshare_amoled_216` and `waveshare_amoled_18`) build clean.

### DIFF
--- a/daemon/claude-usage-daemon.sh
+++ b/daemon/claude-usage-daemon.sh
@@ -12,6 +12,7 @@ REQ_CHAR_UUID="4c41555a-4465-7669-6365-000000000004"
 POLL_INTERVAL=60
 TICK=5
 SAVED_MAC_FILE="$HOME/.config/claude-usage-monitor/ble-address"
+CONFIG_FILE="$HOME/.config/claude-usage-monitor/config"
 REFRESH_FLAG="/tmp/claude-usage-refresh-$$"
 DBUS_DEST="org.bluez"
 NOTIFY_PID=""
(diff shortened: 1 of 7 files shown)


HermannBjorgvin:
I like it more than the current default screen actually, really cool!

devalnor:
I set if to "off" by default but If everyone likes, it we can set it to "auto"

tovbinm:
Ship it!!

HermannBjorgvin:
Merging it. Thank you!

PR #78, opened 2026-06-16
feat: support Enterprise spending-limit accounts

vittorio-nardone:
## Summary

Adds support for **Enterprise spending-limit accounts**, which the daemon previously could not read — they showed `0%` on the device.

Enterprise usage screen

Enterprise accounts return a different set of rate-limit headers than Pro/Max plans. Where Pro/Max expose `anthropic-ratelimit-unified-5h-*` and `-7d-*` (the 5-hour and 7-day usage windows), Enterprise accounts expose `anthropic-ratelimit-unified-overage-*` instead, reflecting a monthly spend-limit model rather than rolling usage windows. The old code read only the 5h/7d headers, found them missing, and defaulted every field to zero — so Enterprise users saw a permanently empty display.

## Daemon changes (macOS, Windows, bash)

- **Account-type detection from headers**: if `anthropic-ratelimit-unified-5h-utilization` is present → Pro/Max; otherwise the account is Enterprise and `overage-utilization` / `overage-reset` are read instead.
- Payload gains an `acct` field (`"pro"` / `"ent"`).
- For Enterprise, the daemon also computes and sends:
  - `tp` — percentage of the billing period elapsed (0–100)
  - `pd` — billing-period length in days
  - `rd` — formatted reset date (e.g. `Jul 1`)

All three daemon implementations were updated to keep behaviour consistent across platforms.

## UI changes (firmware)

The device UI now branches on account type:

- **Pro/Max** — unchanged: the original two boxes (`Current` 5h window + `Weekly` 7d window).
- **Enterprise** — repurposes the two-box layout:
  - **Spending box**: spend percentage (large number + `%`) with the subtitle *"of your monthly budget"*.
  - **Period box**: percentage of the monthly billing period elapsed, plus a **pace indicator** that compares spend against time elapsed and colors the bar and label **green / amber / red** for *under / on / over pace*. The footer reads e.g. *"Under pace - Resets Jul 1"*.

This lets a user see at a glance whether they're spending faster than the billing period is progressing.

## Backward compatibility

Fully backward compatible. Payloads without an `acct` field default to `"pro"`, and the new fields (`tp`, `pd`, `rd`) have safe fallbacks. Pro/Max accounts — and older daemons paired with this firmware — keep the original two-window display unchanged.

## Monthly-period assumption

The rate-limit headers expose only the reset **timestamp**, not the period length. We therefore **assume a calendar-monthly billing period** when computing `tp` / `pd` and when labelling the budget as *monthly*.

This assumption is documented by Anthropic: in the [Claude Enterprise Admin API reference](https://support.claude.com/en/articles/15330651-claude-enterprise-admin-api-reference-guide), the spend-limit `period` field's *"only value today is `monthly`"*. The same doc notes `period` is an open string that may gain other values in the future, so this should be revisited if non-monthly periods appear. The actual `period` value lives in the Admin API (org-level admin key), not in the rate-limit headers this tool uses.

## Testing

- ✅ Tested on **macOS** with a real Enterprise account (live API headers + on-device rendering).
- ✅ Tested on the **Waveshare ESP32-S3-Touch-AMOLED-2.16** board (`waveshare_amoled_216`).
- ⚠️ **Not tested** on other operating systems (Windows, Linux) — the Windows and bash daemons received the equivalent change but could not be exercised here.
- ⚠️ **Not tested** on other boards (`waveshare_amoled_18`); it builds cleanly but was not verified visually.


tovbinm:
@HermannBjorgvin can you please have a look?

HermannBjorgvin:
> @HermannBjorgvin can you please have a look?

Sorry have had limited time to tend to this.

Have you tested this? I don't have an Enterprise account myself my company is still on the teams plan

HermannBjorgvin:
Will merge it if you approve it :)

tovbinm:
OK, I just tried installed it from https://github.com/vittorio-nardone/Clawdmeter/tree/feature/enterprise-account and it works as expected.

tovbinm (approved):

PR #81, opened 2026-06-17
fix(daemon): read OAuth token from Keychain on macOS

alvins82:
## Summary

- Fix the bash daemon's `read_token()` to read from macOS Keychain (`"Claude Code-credentials"` service) instead of `~/.claude/.credentials.json`, which often contains a stale expired token
- Fall back to the credentials file on Linux (unchanged behavior)
- Fix grep pattern to handle optional whitespace around JSON colons (`"accessToken": "..."` vs `"accessToken":"..."`)

## Problem

On macOS, Claude Code stores OAuth credentials in the system Keychain, not in `~/.claude/.credentials.json`. The bash daemon was reading from the file, which holds an old expired token, causing `API HTTP 401: Invalid authentication credentials`. The Python daemon already handled this correctly by reading from Keychain on Darwin.

## Testing

Verified the fixed `read_token()` extracts a valid token from Keychain on macOS and falls back to the file on Linux.

### DIFF
--- a/daemon/claude-usage-daemon.sh
+++ b/daemon/claude-usage-daemon.sh
@@ -21,7 +21,18 @@ log() {
 }
 
 read_token() {
-    grep -o '"accessToken":"[^"]*"' "$HOME/.claude/.credentials.json" | cut -d'"' -f4
+    # macOS: token lives in Keychain (service "Claude Code-credentials").
+    # Linux: token lives in ~/.claude/.credentials.json.
+    if [ "$(uname)" = "Darwin" ]; then
+        local blob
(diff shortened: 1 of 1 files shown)


HermannBjorgvin:
Why are you not just using the MacOS Python daemon? The bash daemon was meant for linux users

alvins82:
Hey @HermannBjorgvin yes i figured that out later but i thought might as well fix this anyway. Can close if not required.

p.s. hope you had a great break.

HermannBjorgvin:
Yes it was a nice break, I am getting caught up on the PR's and issues raised while I was away. Thanks for raising the issue I will close for now since the daemons are per platform specific currently.

PR #83, opened 2026-06-20
fix+feat(daemon): overage-mode usage and per-account-type device screens

skie97:
## Summary

Started as a zero-display bug in overage mode, then — through three-account probing — became a proper per-account screen design.

**The bug.** In overage mode the Anthropic API drops the `5h-utilization` and `7d-utilization` headers entirely. The daemon found nothing, fell back to `0`, and sent `{"s":0,"w":0,"st":"unknown"}` to the device — both bars empty, status unknown.

**`fix(windows-daemon): report overage-mode usage instead of zeros`**
Detects `overage-in-use: true`, marks both windows as 100% / `status="overage"`, prefers the `unified-status` headline (authoritative across all windows), and falls back to `unified-reset` when per-window resets are absent. Also fixes a Python 3.12+ regression where `asyncio.get_event_loop()` raised "no current event loop", silently skipping all 17 poll tests.

**`fix(daemon): port overage-mode handling to macOS and bash daemons`**
Same fix applied to `claude_usage_daemon.py` (macOS) and `claude-usage-daemon.sh` (bash/Linux) so all three daemons report consistently.

**`feat(overage): show real Extra Usage instead of two phantom 100% bars`**
Probing a live overage account showed the API provides `overage-utilization` and `overage-reset`. Daemons now carry these as `o` / `or` on the wire. Both existing Current/Weekly bars retain their real values — the `o`/`or` wire fields are carried for use by the Enterprise screen.

**`feat(overage): per-account-type device screens (Pro/Max Extra Usage + Enterprise)`**
A three-account probe (`probe_headers.py`) showed the discriminator falls straight out of header presence:

| Account | 5h/7d headers | `overage-in-use` | Screen |
|---|---|---|---|
| Pro/Max (normal) | present | absent | **Usage** (unchanged) |
| Pro/Max (overage) | present | `true` | **Extra Usage** — same two real Current/Weekly pills, amber-yellow banner title |
| Enterprise | absent | `true` | **Enterprise** — overage-utilization spend gauge + mini splash strip |

Two new BLE wire fields (`oiu`, `acct`) drive the switch. Enterprise allocates the animation strip from PSRAM (100×100 cells) or falls back to internal SRAM (60×60) on PSRAM-free boards (C6). All three firmware targets build green; both screens verified on hardware.

## Test plan

- [x] Pro/Max normal: existing Usage screen unchanged
- [x] Pro/Max overage: "Extra Usage" amber-yellow banner title, Current/Weekly pills retain real values, second bar shows real overage %
- [x] Enterprise: spend-gauge screen with mini animations
- [x] macOS and bash daemons report same overage payloads as Windows


skie97:
clawd-enterprise
clawd-extra usage

pRieStaKos:
This overrides the clock feature. Is there any other way to display **Enterprise** and **Extra Usage** and keep the clock ?

Suggestion, something like:
**Enterprice**: put the title above usage, as you have space and move usage below
**Extra Usage**: change the usage background or the label "Current" to "Extra Usage"

HermannBjorgvin:
> This overrides the clock feature. Is there any other way to display **Enterprise** and **Extra Usage** and keep the clock ?
> 
> Suggestion, something like: **Enterprice**: put the title above usage, as you have space and move usage below **Extra Usage**: change the usage background or the label "Current" to "Extra Usage"

Suggest changing the "Current" pill on the current session indicator to indicate extra usage instead, i.e. change it from "Current" with a gray background to "Extra usage" with the yellow background and black text.

PR #85, opened 2026-06-21
feat(macos): self-heal stale BLE bond after reflash + harden installer for Python 3.10+

muanton:
## Problem

After a firmware reflash (or the on-device bond-clear gesture), macOS keeps bonding keys that no longer match the ESP32. Every connect then fails with `CBErrorDomain Code=15 "Failed to encrypt the connection"`. CoreBluetooth exposes no unpair API, so the daemon could never recover on its own — the user had to **"Forget This Device"** by hand each time.

Separately, the macOS installer assumed a usable system `python3`, but the daemon uses PEP 604 `X | None` syntax which needs **Python ≥ 3.10** — and macOS ships 3.9, so a fresh install would crash at import on stock systems.

## Changes

**Daemon (`claude_usage_daemon.py`)**
- Detect the `Code=15` / `encrypt` connect error and, on macOS, shell out to `blueutil --unpair` for the bond matching `DEVICE_NAME` (mapped from name via `blueutil --paired`, since the daemon only holds the CoreBluetooth UUID, not the BD_ADDR). The firmware pairs "just works" (no MITM), so the next connect re-bonds silently with no GUI prompt. Mirrors the Linux daemon's `bluetoothctl remove` self-heal.
- `blueutil` talks via IOBluetooth, which needs its **own** Bluetooth TCC grant and **blocks** (rather than erroring) when unauthorized — so every call is bounded by a timeout and a hang is surfaced as a permission problem.
- Degrades gracefully when `blueutil` is absent: logs a manual-fix hint instead of failing.

**Installer (`install-mac.sh`)**
- Probe for a Python ≥ 3.10 interpreter (prefer Homebrew's), and recreate the venv if a previous run built it with an older one.
- Install `blueutil` via Homebrew (best-effort) and prime its Bluetooth permission behind a bounded wait, so the one-time prompt fires at install time rather than mid-recovery.

## Notes / testing

- `python -m py_compile` and `bash -n install-mac.sh` both clean.
- The self-heal path triggers only on a real stale-bond `Code=15` and is independent of the unrelated start_notify hang fixed in #84.

### DIFF
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -11,6 +11,7 @@ import getpass
 import json
 import os
 import re
+import shutil
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
Thanks for the fix!

PR #91, opened 2026-06-23
fix(daemon): bound client.connect() so a half-open BLE link can't wedge the daemon (#90)

### LINKED ISSUE #90: macOS/Linux daemon: unbounded BLE connect can wedge the daemon, same bug class as start_notify (#84)
## Summary

#84 (merged via #59) bounded `start_notify` so a half-open BLE link can't wedge the macOS/Linux daemon between **"Connected"** and the first poll. But the **preceding** await — `client.connect()` itself, which runs **before** "Connected" — is still unbounded. It has the same half-open-link failure shape #84 described, just one call site earlier. So the daemon can still hang forever, this time *before* it ever logs "Connected".

This is a follow-up sibling of #84 (same bug class, different call site), exactly as #89 is for the Windows daemon's `start_notify`.

## Where

`daemon/claude_usage_daemon.py`, `connect_and_run()`:

https://github.com/HermannBjorgvin/Clawdmeter/blob/52b23c8/daemon/claude_usage_daemon.py#L450-L459

```python
log(f"Connecting to {display}...")
client = BleakClient(target)
try:
    await client.connect()          # <-- unbounded; no asyncio.wait_for
except (BleakError, asyncio.TimeoutError) as e:
    log(f"Connection failed: {e}")
    ...
    return False
```

Contrast with the `start_notify` site #84 already fixed a few lines down, which *is* bounded:

https://github.com/HermannBjorgvin/Clawdmeter/blob/52b23c8/daemon/claude_usage_daemon.py#L335-L343

## Behavior

```gherkin
Scenario: A Bluetooth connect that never completes is abandoned so the daemon keeps serving fresh data
  Given the firmware is advertising and the OS reports the physical HID link as connected
  And the GATT connect handshake never completes (a half-open link)
  When the daemon attempts to connect to the device
  Then the connect attempt is abandoned after a bounded timeout
  And the daemon logs the failure, drops the cached address, and rescans
  And no single connect attempt blocks the daemon for longer than that timeout
  # CURRENTLY: await client.connect() has no timeout, so the daemon hangs indefinitely at
  # "Connecting..." — never reaching "Connected" — and the device shows stale data until a manual restart
```

## Evidence (observed in the wild)

The daemon hung for **~13 hours** with the last log line being a `Connecting...` that never reached `Connected`:

```
[17:23:04] Found: 736D74B8-2ADB-6219-80DE-31AE4290F9BE
[17:23:04] Connecting to 736D74B8-2ADB-6219-80DE-31AE4290F9BE...
            <no further output for ~13h>
```

- Process alive but asleep (state `S`), ~0% CPU, ~1:19 of CPU time over 21h elapsed.
- Device kept displaying the **last successfully delivered** snapshot (an expired 5h window), so utilization and the reset countdown were badly wrong with **no error and no recovery**.
- `launchctl kickstart -k` unblocked it instantly — the `SIGTERM` cancelled the stuck `connect()` and a fresh start reconnected and resumed `Sending:` within ~1s.

This is the connect-step analogue of the `start_notify` hang #84 reproduced (29h hung process, 0% CPU, no log past `Connected`).

## Why `connect(timeout=...)` isn't enough

`BleakClient.connect()` takes a `timeout` (default 10s), but on the CoreBluetooth backend that timeout governs peripheral *discovery*, not the `connectPeripheral` handshake — CoreBluetooth's connect has no native timeout, so a half-open link can stall the await indefinitely. The empirical 13h hang confirms the default did not fire here. An explicit `asyncio.wait_for` is needed, exactly as #84 concluded for `start_notify`.

## Suggested fix

Mirror #84 — wrap the connect in `asyncio.wait_for(...)`. Note the existing `except` clause **already catches `asyncio.TimeoutError`**, so the change is essentially one line and the failure already routes into the existing invalidate-cache → back off → rescan cycle:

```python
try:
    await asyncio.wait_for(client.connect(), timeout=20)
except (BleakError, asyncio.TimeoutError) as e:
    log(f"Connection failed: {e}")
    ...
    return False
```

Unlike `start_notify` (optional — log and proceed), a failed connect must `return False` so the caller re-scans; the current handler already does exactly that.

## Environment

- macOS, `bleak 3.0.2`, daemon at commit `52b23c8` (current `main`).
- **Not affected:** the Bash daemon (`claude-usage-daemon.sh`) connects via `bluetoothctl`/`busctl`, which are bounded by their own method-call timeouts (same reasoning as #89's "Not affected" note).

I'm happy to open a PR with the one-line `wait_for` wrap if that's welcome.

ryanrozich:
Closes #90.

## What

`#84` (merged via #59) bounded `start_notify` so a half-open BLE link can't wedge the macOS/Linux daemon between **"Connected"** and the first poll. The **preceding** await — `client.connect()` itself — was still unbounded, so the daemon can hang one call site earlier, **before** it ever logs "Connected".

This wraps the connect in `asyncio.wait_for(client.connect(), timeout=CONNECT_TIMEOUT)`, mirroring #84.

```diff
-        await client.connect()
+        await asyncio.wait_for(client.connect(), timeout=CONNECT_TIMEOUT)
```

The existing `except (BleakError, asyncio.TimeoutError)` already treats a timeout as a connection failure and `return False`s, so a stalled connect now routes straight into the current invalidate-cache → back off → rescan cycle instead of blocking forever. New constant `CONNECT_TIMEOUT = 20.0` sits beside `SCAN_TIMEOUT`.

## Why

On macOS the firmware advertises as a BLE HID keyboard, so the OS auto-connects the physical link and the daemon rides it. If CoreBluetooth hands back a half-open peripheral whose GATT connect handshake never completes, the await hangs — and `BleakClient`'s own `timeout` governs *discovery*, not `connectPeripheral`, so it doesn't fire.

Observed in the wild: the daemon hung **~13h** with the last log line a `Connecting...` that never reached `Connected` — process alive but asleep, ~0% CPU, device stuck showing an expired 5h window with no error and no recovery until a manual `launchctl kickstart -k`. This is the connect-step analogue of the `start_notify` hang #84 reproduced.

## Testing

- `python -m py_compile daemon/claude_usage_daemon.py` — clean.
- A `wait_for` timeout raises `asyncio.TimeoutError`, which the unchanged handler already catches → `return False` → caller rescans (same recovery path #84 relies on).
- 20s is comfortably above a healthy connect (~1–6s in the logs) while capping the worst case.

Scoped intentionally to the connect hang; happy to adjust the timeout value or wording.

### DIFF
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -30,6 +30,7 @@ REQ_CHAR_UUID = "4c41555a-4465-7669-6365-000000000004"
 POLL_INTERVAL = 60
 TICK = 5
 SCAN_TIMEOUT = 8.0
+CONNECT_TIMEOUT = 20.0
 
 # macOS: token lives in Keychain (service "Claude Code-credentials").
 # Linux: token lives in ~/.claude/.credentials.json.
@@ -450,7 +451,16 @@ async def connect_and_run(target, stop_event: asyncio.Event) -> bool:
     log(f"Connecting to {display}...")
     client = BleakClient(target)
     try:
-        await client.connect()
+        # Bound the connect the same way #84 bounded the refresh subscribe.
+        # On macOS the OS auto-connects the firmware's HID link, so
+        # CoreBluetooth can hand us a half-open peripheral whose GATT connect
+        # handshake never completes. BleakClient's own timeout governs
+        # discovery, not connectPeripheral, so an unbounded await here wedges
+        # the single-threaded daemon forever at "Connecting..." (observed ~13h,
+        # device stuck on stale data). wait_for raises TimeoutError, which the
+        # handler below already treats as a connection failure -> drop the
+        # cached address and rescan.
+        await asyncio.wait_for(client.connect(), timeout=CONNECT_TIMEOUT)
     except (BleakError, asyncio.TimeoutError) as e:
         log(f"Connection failed: {e}")
         if sys.platform == "darwin" and _is_encryption_error(e):


HermannBjorgvin (approved):

HermannBjorgvin:
Brilliant! Thank you so much. I need to get a Mac one of these days.

PR #99, opened 2026-07-01
system-peripheral-only by default: lock daemon + board to one machine

pRieStaKos:
## What & why

This PR started by adding an opt-in `system_peripheral_only` flag. It's now the
permanent, only behavior: the daemon **and** the board both lock to a single
machine, so the display can't be grabbed by — or rotate to — another machine's
daemon that's in BLE range.

### Daemon (macOS / Linux / Windows)

- The `system_peripheral_only` config flag is **removed** — pinned-only is now
  the default and only mode.
- **All scan-by-name fallback removed.** Each daemon only ever connects to the
  peripheral the OS already holds: macOS system-connected peripheral, Linux
  bluez paired-or-connected entry, Windows bonded PnP device. The device is a
  bonded BLE HID keyboard you pair once, so this matches the documented setup.
- `config_dirs` multi-plan rotation is **unchanged** — only the connection half
  of the old flag became the default.
- Linux: no longer `bluetoothctl remove` on a failed connect (that would unpair
  the pinned device and strand the daemon).
- Installers drop the `system_peripheral_only` prompt; tests updated to assert
  no-scan is the default; README pairing wording fixed.

### Firmware (`ble.cpp`) — single-owner lock

The daemon change stops *your* daemon grabbing the wrong board, but it can't
stop *another* machine's daemon writing to *your* board — the board previously
accepted usage data from any connected central. That's now enforced on-device:

- Locks to one **owner** = the identity address of the bonded machine,
  persisted in NVS across reboots.
- Usage writes are accepted only over a **bonded + encrypted** link and only
  from the owner; writes from any other central are ignored (and logged).
- A second machine that pairs is un-bonded and disconnected; the owner can
  always re-pair, so it can't be locked out. Non-owner bonds are pruned on boot.
- Hold-power ~3s clears the bond + owner to hand the board to another machine.
  `MAX_CONNECTIONS` stays 2, so Windows' separate HID + GATT links still work.

## Verification

- Daemon test suite passes per-file (macOS multidir, Windows bonded / reconnect
  / tray).
- Firmware builds for both `waveshare_amoled_216` and `waveshare_amoled_18`.
- Live: with a second machine's daemon on and in range, the board shows only
  the owner's account and no longer rotates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The board env name is required. Run `./flash-mac.sh` with no args to see the ava
 
 ### Pair the device
 
-After flashing, open **System Settings → Bluetooth** and click *Connect* next to "Clawdmeter". The daemon will discover it on its next scan (~30 s).
+After flashing, open **System Settings → Bluetooth** and click *Connect* next to "Clawdmeter". The daemon only ever connects to the peripheral this Mac is paired/connected to — it never scans for a nearby device — so once it's connected here the daemon picks it up on its next poll (~60 s).
 
 ### Install the daemon
 
--- a/daemon/claude-usage-daemon.sh
+++ b/daemon/claude-usage-daemon.sh
@@ -95,24 +95,6 @@ read_clock_setting() {
     esac
 }
 
-# Read the `system_peripheral_only` option. Echoes one of: on|off (default off).
-# When on, the daemon connects only to a device the system already knows (paired/
-# connected — see find_system_device_mac, never a name scan) and shows only this
-# machine's own account (~/.claude), disabling multi-plan rotation.
-read_system_only_setting() {
-    local val=""
-    if [ -f "$CONFIG_FILE" ]; then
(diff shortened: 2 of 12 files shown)


HermannBjorgvin:
It's a good point I didn't consider in the scanning for named devices that you might have more than one nearby.

In my opinion this could just be the default and we can remove the configuration option and drop all the fallback scanning for named devices. What do you think?

pRieStaKos:
> It's a good point I didn't consider in the scanning for named devices that you might have more than one nearby.
> 
> In my opinion this could just be the default and we can remove the configuration option and drop all the fallback scanning for named devices. What do you think?

I noticed this from the begging, as I bought two boards and got Usage rotation for different accounts.

Let me switch to default and isolate to system_peripheral_only. You are right. I think it's OK

HermannBjorgvin:
Yeah, simplifies the logic as well to get rid of the fallback scan, thanks for bringing this up.

pRieStaKos:
@HermannBjorgvin Updated to default

HermannBjorgvin:
Thanks @pRieStaKos !

PR #103, opened 2026-07-12
Opt-in usage log + HTML report tool

leonardovalim:
## What

Adds an **opt-in** usage log and a small standalone tool to turn it into an HTML report.

- New config option `usage_log = on|off` (default **off** — nothing is written unless you opt in).
- When on, both Python daemons (Windows + macOS) append one JSON-lines record per poll to `usage_log.jsonl` in the config dir, capturing the session/weekly rate-limit utilization.
- `tools/usage_report.py` reads that log and renders a self-contained, **dependency-free** HTML report (stdlib only): summary cards, session/weekly line charts, and a day-by-hour heatmap of when usage peaks.

## Why

The daemon already polls utilization every ~60s but only forwards the latest snapshot to the device. Persisting it (when the user opts in) makes it easy to look back over time — e.g. "when do I hit the 5h limit most?" — with no external service.

## Notes

- Opt-in + best-effort: a logging failure never disturbs the poll loop; the file is only touched when `usage_log = on`.
- No third-party deps in the report tool (pure stdlib + inline SVG), so it runs anywhere Python 3.9+ does.
- Scoped to the two Python daemons; the bash daemon could get the same hook if you'd like.
- Tests in `daemon/tests/test_usage_log.py` cover the config gate and field selection.

Happy to adjust naming/scope/placement to fit the project's conventions.

### DIFF
--- a/daemon/claude_usage_daemon.py
+++ b/daemon/claude_usage_daemon.py
@@ -40,6 +40,10 @@ DEFAULT_CONFIG_DIR = Path.home() / ".claude"
 SAVED_ADDR_FILE = Path.home() / ".config" / "claude-usage-monitor" / "ble-address"
 CONFIG_FILE = Path.home() / ".config" / "claude-usage-monitor" / "config"
 
+# Opt-in usage log: one JSON-lines record per poll for tools/usage_report.py.
+# Only written when `usage_log = on` in the config; append-only, self-describing.
+USAGE_LOG_FILE = Path.home() / ".config" / "claude-usage-monitor" / "usage_log.jsonl"
+
 API_URL = "https://api.anthropic.com/v1/messages"
 API_HEADERS_TEMPLATE = {
     "anthropic-version": "2023-06-01",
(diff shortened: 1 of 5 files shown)


HermannBjorgvin:
I'm feeling like this is out of scope for this project anyway I don't want to take ownership of maintaining this feature.

Sorry not the news you wanted but I'm trying to keep this project singularly focused.

I'm more open to letting the daemon connect to a third party application if anyone makes a great usage viewer app for example and they take on the maintaining of that because even though the cost of code is close to zero now the cost to maintain a project is so non zero so I need to limit the scope of what I want to maintain personally.

leonardovalim:
Totally fair — thanks for taking the time to explain the reasoning, it makes a lot of sense. Keeping the maintenance surface small is the right call for a project like this, and the “writing code is cheap, maintaining it isn’t” point really resonates.I’ll keep the report/logging as a separate tool on my side. Appreciate the thoughtful response and I’ll aim future contributions at bug fixes that stay inside your scope. Thanks for the great project! 🙏Em 12 de jul. de 2026, à(s) 10:51, Hermann Björgvin ***@***.***> escreveu:﻿
Closed #103.

PR #105, opened 2026-07-13
fix(amoled-18): amp_enable must not drive EXIO2 — it kills FT3168 touch

jkbdco:
## Summary

On the ESP32-S3-Touch-AMOLED-1.8 (SH8601 + FT3168 revision), touch has been dead from boot since the chime landed (2f94248). Root cause: the board's `amp_enable` hook drives **both** amp-enable candidates — GPIO 46 and XCA9554 **EXIO2** — on the assumption that "driving the unused one is harmless." It isn't: **pulling EXIO2 low takes the FT3168 touch controller off the I2C bus** until the line goes high again. Since `chime_init()` calls `amp_enable(false)` at boot to park the amp off, EXIO2 stayed low and touch never worked.

Fix: drive GPIO 46 only. EXIO2 is parked HIGH by `io_expander_init()` and must stay that way (this is also what the pre-chime firmware always did).

## How it was found (hardware bisection on a real 1.8)

The failure is deceptive: once the FT3168 vanishes, IDF's `i2c_master` returns `ESP_ERR_INVALID_STATE` on every transaction to 0x38, which looks like a wedged I2C driver rather than an unpowered chip. Bisection results:

| Build | Touch probe |
|---|---|
| chime fully skipped | ✅ `FT3168 ID=0x03` |
| I2S up, codec skipped (real pins) | ❌ |
| I2S on parked GPIOs 17/18/13 | ❌ |
| **amp lines only, no I2S/codec** | ❌ ← poison confirmed |
| GPIO 46 only, EXIO2 untouched | ✅ |
| **this PR (full chime + fix)** | ✅ |

A probe matrix after the failure showed 0x18/0x20/0x34 answering normally on both the raw HAL and `Wire` paths while only 0x38 failed — i.e. the bus and driver were fine, the chip was gone.

## Verified on hardware (SH8601 + FT3168 rev)

- Boot log clean: `chime: ES8311 ready` **and** `Touch FT3168 ID=0x03`, zero I2C errors
- Tap-to-switch UI and PWR button work with the chime engine fully initialized
- Chime audibly plays through the speaker via the real BLE path (session-reset payload with `"c":1`), so the "untested on hardware" caveat in sound.cpp is now covered
- BLE bond + owner persistence across power cycles unaffected

CLAUDE.md gets a session-highlights entry documenting the EXIO2 behavior so future sessions don't re-chase the I2S red herring.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -142,6 +142,7 @@ See `~/.claude/projects/.../memory/` files for persistent context (user is an em
 
 ## Recent session highlights
 
+- **AMOLED-1.8 chime verified on hardware + EXIO2 touch-kill fix (2026-07-13).** The 1.8's `amp_enable` hook drove both GPIO 46 and XCA9554 EXIO2 ("the unused one is harmless") — but pulling EXIO2 low takes the FT3168 off the I2C bus (chip stops ACKing; IDF reports it as `ESP_ERR_INVALID_STATE`, which reads like a driver wedge and cost a long I2S red-herring chase). Amp enable is GPIO 46 only; EXIO2 must stay HIGH. Chime, touch, buttons, and BLE bond persistence all verified on a real 1.8.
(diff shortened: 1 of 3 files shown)


pRieStaKos:
Interesting...

HermannBjorgvin:
Thanks for contributing as fix for this and for the analysis.

HermannBjorgvin (approved):

Issue #89, opened 2026-06-21
Windows daemon: unbounded start_notify can wedge like the macOS daemon (#84)


HermannBjorgvin:
## Summary

The fix in #84 bounds `start_notify` in `claude_usage_daemon.py` (Linux/macOS) so a half-open BLE link can't wedge the daemon. The **Windows daemon (`claude_usage_daemon_windows.py`) has the same bug class and is not covered by that PR.**

This is a follow-up, **not blocking for #84.**

## Details

`setup_refresh_subscription()` does an unbounded `await self.client.start_notify(...)`:

https://github.com/HermannBjorgvin/Clawdmeter/blob/main/daemon/claude_usage_daemon_windows.py#L185

It sits on the path between connect and the poll loop (`await session.setup_refresh_subscription()` ~line 371), identical in structure to the macOS daemon. The existing `except (BleakError, ValueError, OSError)` does **not** help: an exception handler only catches a *raised* error, never an `await` that hangs forever. If WinRT's CCCD-write confirmation never arrives on a half-open link (same failure shape as the CoreBluetooth hang #84 fixed), the single-process daemon wedges between "Connected" and the first poll — stale data, no error, no recovery until manual restart.

## Fix

Mirror #84: wrap the subscribe in `asyncio.wait_for(..., timeout=10)` and add an `except asyncio.TimeoutError` that logs and proceeds (the refresh subscription is optional; the 60s poll loop works without it). Keep the existing `OSError` in the catch list since WinRT can surface failures as raw `OSError`.

## Not affected

The Bash daemon (`claude-usage-daemon.sh`) is **not** vulnerable: it subscribes via `busctl call ... StartNotify`, which is bounded by sd-bus's default ~25s method-call timeout and whose result is discarded, so it can't block indefinitely.

Ref: #84

Issue #88, opened 2026-06-21
Windows daemon: no self-heal for stale BLE bond after firmware reflash


HermannBjorgvin:
🤖 AI Generated:

## Problem

After a firmware reflash (or the on-device bond-clear gesture), the host keeps BLE bonding keys that no longer match the ESP32, so every reconnect fails to encrypt the link. PR #85 added self-heal for **macOS**, and the **Linux** bash daemon already self-heals via `bluetoothctl remove`. The **Windows** daemon (`daemon/claude_usage_daemon_windows.py`) has no equivalent recovery — this is a gap with the same root cause.

## Why Windows is affected

The same chain that bites macOS applies to Windows, and it is documented:

1. Firmware enables bonding and advertises an HID keyboard service that requires encryption (`firmware/src/ble.cpp:151`, `setSecurityAuth(true, false, true)`).
2. Windows pairing is **mandatory and explicit** — the setup requires pairing via Settings → Add device ("This is required — the device is a bonded BLE HID keyboard", `daemon/README-windows.md:17,38-50`). So Windows definitely stores a bond.
3. A reflash invalidates the ESP32 keys, but Windows keeps the stale bond → reconnect fails to encrypt (the `Code=15`-equivalent failure that PR #85 heals on macOS).
4. The Windows daemon has no pair/unpair/encryption handling. Its only stale-state mitigation is `use_cached_services=False` (`daemon/claude_usage_daemon_windows.py:335`), which addresses the GATT **service cache** — a different problem. The documented recovery is the manual **Settings → Remove device** (`README-windows.md:57`), i.e. exactly the hand-forget dance macOS users had to do before #85.

## Why this is not a straight port of the macOS fix

- On macOS, "just works" pairing means after `blueutil --unpair` the next connect **re-bonds silently**, so unpair alone restores the device.
- On Windows, the daemon’s **data** characteristics use plain `WRITE`/`NOTIFY` with no `_ENC` (`ble.cpp:185-200`), so the data connection could reconnect unbonded — but the **HID keyboard buttons** require a Windows-level bond. After a programmatic `DeviceInformation.Pairing.UnpairAsync()`, fully restoring the device (including buttons) likely also needs a `PairAsync(ConfirmOnly)` re-pair.

## Proposed approach

[...]

PR #106, opened 2026-07-13
fix(windows): end the BLE reconnect churn — widen supervision timeout (2s→6s) + crash-proof the daemon

tacticalBeard-dev:
## Problem

On native Windows 11, the device churns through BLE connections — the daemon log shows hundreds of `Connecting → Connected → Refresh subscription unavailable: [WinError -2147023673] The operation was canceled by the user → Device disconnected` cycles per day, most connections dying **sub-second** during the CCCD subscribe, with occasional ~30s survivals. Underneath, every ~15–45 min a bleak WinRT fault escaped the daemon's exception handling and **permanently killed the polling thread** — the tray showed "Error: daemon crashed: AssertionError" and the device sat frozen on stale data until manual relaunch.

### Environment where this reproduces

- Native Windows 11 (no WSL), Python 3.12, bleak 3.0.2, winrt 3.2.1
- Waveshare ESP32-S3-Touch-AMOLED-2.16 (the README's reference board)
- Device **within 2–3 feet of the host, no line-of-sight obstruction** — this is not a range problem
- Other BLE devices on the same host work fine at 20 ft through walls

The trigger is environmental RF sensitivity, not distance: with the stock 2s Windows supervision timeout, simply **picking the device up and turning it** near the host reliably killed the link (reason=520 in the serial log), and in normal stationary desk use it dropped periodically for no visible reason. If your desk happens to have clean RF and a friendly geometry you will never see this — which is likely why it hasn't been reported: macOS centrals already run ~30ms/6s (immune by default), and Windows users without local interference never get close to a 2s gap. For anyone with unknown interference sources, a metal-heavy desk, a rear-panel Bluetooth adapter, or just unlucky antenna orientation, the device is effectively unusable on Windows without this fix — and nothing in the UI or logs tells them why.

Two distinct layers, both fixed here:

1. **Firmware/link layer:** every single disconnect in serial captures was `reason=520` (0x208 = **supervision timeout**). Windows as a BLE central imposes `interval=15ms latency=0 supervision-timeout=2s` on every bonded reconnect, and clamps back to 2s whenever an app GATT session goes active — while it happily runs **9.6s** when only its HID driver holds the link (all values captured on hardware via new conn-params logging). A 2s window is tight enough that ordinary orientation changes near the host — deep RF nulls from the board's PCB antenna being shadowed — kill the link during normal handling.

2. **Daemon/host layer:** bleak 3.0.2's WinRT backend raises a **bare `AssertionError`** (`FutureLike.result`, a completion race provoked by the flaky link) and raw `OSError` during connect/service-discovery — neither wrapped as `BleakError`, so they escaped `connect_and_run`'s `except (BleakError, asyncio.TimeoutError)` and killed the thread. A separate one-shot crash: `datetime.timestamp()` on Windows raises `OSError(22)` for a garbage/zero `anthropic-ratelimit-unified-overage-reset` header (`_billing_period_info`). And the tray's `_run_daemon` had no restart path — any crash was permanent.

## How it was found

Serial capture (`pio device monitor`) during live churn, with temporary conn-params + loop-timing instrumentation:

| Hypothesis | Verdict | Evidence |
|---|---|---|
| Owner-lock force-disconnecting on unresolved RPA (leading theory going in) | ❌ refuted | 0 `rejecting non-owner` lines across all captures; auth always `bonded=1 enc=1`; peer identity address stable across reconnects |
| CPU starvation blocking the BLE stack | ❌ refuted | every `loop()` tick timed: max ~119ms (LVGL splash render 82ms steady); NimBLE host runs in its own task |
| Stale/asymmetric bond state | ❌ refuted | full unpair + re-pair → churn resumed within ~6 min |
| Supervision timeout too tight | ✅ confirmed | all disconnects `reason=520`; Windows negotiates 2s (`connparams itvl=12(15.00ms) lat=0 timeout=200(2000ms)` on every bonded reconnect); moving the device 2 ft from the host triggered 520s on demand |

What didn't work: requesting relaxed params **immediately** in `onAuthenticationComplete` — Windows answered the update transaction but kept its 2s. (Windows also has a documented late-instant bug in its param-update path — Microsoft Q&A 848142 — that can itself cause 0x08 disconnects, so param-update traffic is kept to one deferred request.)

## Fix

**Firmware** (shared `ble.cpp` + all four board envs):

- **PPCP characteristic** (GAP 0x2A04) via `MYNEWT_VAL_BLE_SVC_GAP_PPCP_*` build flags: interval 15–30ms, latency 0, supervision timeout 6s — the accessory's declared steady-state preference per Microsoft's Bluetooth accessory guidelines (range includes 15/30ms as they require). Windows caches GATT per bond, so PPCP changes are seen after a re-pair.
- **One-shot deferred connection-parameter request:** when a connection runs at a timeout below 6s (armed in `onAuthenticationComplete` for bonded reconnects that *start* clamped, and in `onConnParamsUpdate` for post-pairing clamps), `ble_tick()` sends a single LL request ~2s later for 15–30ms / 0 / 6s. Deferred so it can't race the central's own in-flight update; a spent-handle guard makes it one request per connection, so a central that re-clamps can't trigger an update war.
- Negotiated conn params are now logged on connect and on every update — `reason=520` is only diagnosable next to the negotiated timeout.

The resulting steady state (30ms/6s) is exactly what macOS centrals already use — this brings the Windows link to parity with the platform where nobody reports churn. The single-owner security lock is deliberately untouched.

**Daemon** (`claude_usage_daemon_windows.py`, `tray_windows.py`):

- `connect_and_run` treats bleak's bare `AssertionError`/`OSError` as a normal failed attempt feeding the existing retry loop (and swallows the `disconnect()` teardown assert on an already-gone peer).
- `_billing_period_info` guards the Windows `timestamp()`/`fromtimestamp()` range errors and degrades to the safe default payload.
- `_run_daemon` is now a supervisor: crash → log traceback → tray error state → restart with capped backoff (2s→30s). Quit sets a `_quit_requested` event *before* signalling stop, so a clean shutdown is never resurrected and the backoff wait is Quit-interruptible.

## Testing

- ✅ `pytest daemon/tests/`: **114 passed, 2 skipped** (Windows 11, Python 3.12). Includes 2 test-harness fixes found during verification: `test_windows_poll._run()` now uses a fresh event loop per call (the old `asyncio.get_event_loop()` returned a closed loop once another test file's `asyncio.run()` closed the process default — 17 order-dependent false failures), and the two Linux/WSL-warning tests `skipif` on win32 (they assert a warning that is correctly suppressed on native Windows).
- ✅ All four board envs compile: `waveshare_amoled_216`, `waveshare_amoled_18`, `waveshare_amoled_216_c6`, `waveshare_amoled_18_c6`.
- ✅ Hardware-verified on **Waveshare ESP32-S3-Touch-AMOLED-2.16 + Windows 11** (bleak 3.0.2, winrt 3.2.1):
  - Windows now *initiates* reconnects at `itvl=48(60ms)` fresh-pair and settles at `itvl=24(30.00ms) lat=0 timeout=600(6000ms)` — captured across many reconnect cycles.
  - A forced-null disconnect (device held in an RF-shadowed orientation) measurably takes the **full 6s** instead of 2s.
  - Field test in the reproducing environment (device 2–3 ft from host, known-flaky RF): pre-fix, gentle handling killed the link in ≤2s and the daemon log accumulated 393 connection attempts / 163 losses over 2 days. Post-fix, on battery in its usual desk spot: **~25 min stable, a single 7-second self-healed blip** (disconnect → reconnected → fresh data in 7s); survives being carried and rotated around the host; killing the link now requires deliberately holding it in an antenna-shadowed orientation for the full 6s window. The tray survived hundreds of forced flaps (including a zombie-link storm) with zero crashes.
- ⚠️ Not hardware-tested: the other three boards (compile-only) and macOS/Linux centrals (reasoned: macOS already runs ≥6s timeouts, so the pushback never arms; PPCP is standard and advisory).

## Out of scope / related

- #16 specced a firmware supervision timeout (item B) that was never landed — this PR completes that item with measured evidence.
- #88 (Windows stale-bond self-heal) and #89 (unbounded `start_notify`) are adjacent Windows-daemon issues, not addressed here.
- Follow-up feature proposal: on-device BLE signal-strength bars (RSSI), which would have made this diagnosis visible at a glance.

### DIFF
--- a/daemon/claude_usage_daemon_windows.py
+++ b/daemon/claude_usage_daemon_windows.py
@@ -267,12 +267,21 @@ def _billing_period_info(now: float, reset_ts: str) -> dict:
         # raises OSError for pre-1970 dates on Windows — taking the whole poll
         # loop down. Bail out to the neutral default instead.
         return {"tp": 0, "pd": 30, "rd": ""}
-    dt_end = datetime.datetime.fromtimestamp(period_end)
-    prev_month = dt_end.month - 1 or 12
-    prev_year = dt_end.year if dt_end.month > 1 else dt_end.year - 1
-    prev_day = min(dt_end.day, calendar.monthrange(prev_year, prev_month)[1])
-    dt_start = dt_end.replace(year=prev_year, month=prev_month, day=prev_day)
-    period_start = dt_start.timestamp()
+    try:
+        dt_end = datetime.datetime.fromtimestamp(period_end)
+        prev_month = dt_end.month - 1 or 12
+        prev_year = dt_end.year if dt_end.month > 1 else dt_end.year - 1
+        prev_day = min(dt_end.day, calendar.monthrange(prev_year, prev_month)[1])
+        dt_start = dt_end.replace(year=prev_year, month=prev_month, day=prev_day)
+        period_start = dt_start.timestamp()
+    except (OSError, OverflowError, ValueError):
+        # Belt-and-braces beyond the <= 0 guard above (#104): Windows
+        # datetime.timestamp()/fromtimestamp() also raise OSError(22)/
+        # OverflowError/ValueError for out-of-range NON-zero values (e.g. a
+        # far-future "99999999999999" header, which overflows fromtimestamp).
+        # Garbage must never crash the daemon thread — degrade to the safe
+        # default instead (field report: OSError(22) killed the poll loop).
+        return {"tp": 0, "pd": 30, "rd": ""}
     period_len = period_end - period_start
     if period_len <= 0:
         return {"tp": 0, "pd": 30, "rd": ""}
@@ -540,8 +549,15 @@ async def connect_and_run(device, stop_event: asyncio.Event, tray_state=None) ->
         )
         try:
             await client.connect()
-        except (BleakError, asyncio.TimeoutError) as e:
-            log(f"Connection attempt {attempt + 1}/{CONNECT_RETRIES} failed: {e}")
+        except (BleakError, OSError, asyncio.TimeoutError, AssertionError) as e:
+            # WinRT service discovery inside connect() can surface a raw OSError
+            # (WinError) or even a bare AssertionError from bleak's FutureLike
+            # (assert self._result) when the peer drops the link mid-discovery —
+            # neither is wrapped as BleakError. Treat them as a normal failed
+            # attempt so the D-01 retry loop handles them, instead of letting an
+            # uncaught exception kill the daemon thread (the "daemon crashed"
+            # tray toast + silent polling stop, field report).
+            log(f"Connection attempt {attempt + 1}/{CONNECT_RETRIES} failed: {type(e).__name__}: {e}")
             try:
                 await client.disconnect()
             except BleakError:
@@ -625,7 +641,10 @@ async def connect_and_run(device, stop_event: asyncio.Event, tray_state=None) ->
         # so swallow both; the link tears down regardless once we exit.
         try:
             await client.disconnect()
-        except (BleakError, OSError):
+        except (BleakError, OSError, AssertionError):
+            # bleak's WinRT disconnect() also has bare asserts (e.g. assert char
+            # while tearing down notifications on an already-gone peer); swallow
+            # it too — the link tears down regardless once we exit.
             pass
 
     log("Device disconnected" if not stop_event.is_set() else "Stopping")
--- a/daemon/tests/test_windows_poll.py
+++ b/daemon/tests/test_windows_poll.py
@@ -33,8 +33,19 @@ def _make_mock_response(status_code=200, headers=None):
 
 
 def _run(coro):
-    """Run a coroutine synchronously for synchronous test functions."""
-    return asyncio.get_event_loop().run_until_complete(coro)
+    """Run a coroutine synchronously for synchronous test functions.
+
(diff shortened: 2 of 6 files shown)


tacticalBeard-dev:
Rebased onto main to resolve a conflict with #104, which independently fixed the same `_billing_period_info` OSError(22) crash while this branch was in flight.

**Resolution:** kept both layers, since they cover different inputs:

- #104's `period_end <= 0` early return stays as the readable fast path for the common trigger (absent overage-reset header → `"0"` → 1969 after the month step-back).
- This PR's `try/except (OSError, OverflowError, ValueError)` around the datetime arithmetic remains as a catch-all for out-of-range **non-zero** values (e.g. a far-future `"99999999999999"` header still raises through `fromtimestamp` on Windows and would have killed the poll loop past #104's guard alone).

Verified after the rebase: full suite 114 passed / 2 skipped on Windows 11 + Python 3.12, and direct checks of `_billing_period_info` against `'0'`, `'-1'`, `'99999999999999'`, `'not-a-number'` (all → safe default) and a valid future timestamp (correct period computed). No other files conflicted; the firmware commit rebased untouched.

HermannBjorgvin:
Thanks for the fix! Appreciate it!

PR #108, opened 2026-07-15
Add SpotPear ESP32-S3-LCD-1.54 (MUMA) board port + tiny 240x240 layout

bluh-jcrubioa:
## What

Adds an (unofficial, community) board port for the **SpotPear ESP32-S3-LCD-1.54-MUMA** (N16R8) — the project's **first non-AMOLED panel**: a plain 4-wire SPI **ST7789 240×240 IPS**. Build env: `spotpear_s3_154`.

## Hardware

- **Display:** ST7789 240×240 via `Arduino_ESP32SPI` + `Arduino_ST7789` @ 40 MHz (RGB order, INVON for the IPS). Backlight is **LEDC PWM on an active-low GPIO** (duty inverted; the panel has no brightness command).
- **Touch:** CST816 @ `0x15`, reusing the same inline FocalTech-style reader as the 1.8.
- **Power / IMU / sound:** stubbed (`BOARD_HAS_BATTERY/IMU/SOUND=0`; no PMU). No rotation.
- **Pinout** verified from the xiaozhi-esp32 `sp-esp32-s3-1.54-muma` board config plus an on-hardware I2C scan — this board does **not** follow the Waveshare 1.54 pinmap despite SpotPear's wiki linking to it. (LCD SCLK=4, MOSI=2, CS=5, DC=47, RST=38; BL=42 active-low; touch SDA=11/SCL=7, INT=12, RST=6; BOOT=0.)

The env mirrors `waveshare_amoled_18` (S3, 16 MB flash, `qio_opi` OPI PSRAM) minus the unused SensorLib/XPowersLib deps.

## UI: new "tiny" layout breakpoint (H < 300)

240×240 previously fell into the *compact* (368×448) branch and overflowed badly. This adds a dedicated tiny branch and finishes the parameterization so it renders cleanly:

- **Parameterized in-panel fonts** (pct / pill / reset) + bar height. This also fixes a pro-account update path that hard-set the session `%` to `styrene_48` regardless of layout — so the 5h number rendered huge and clipped.
- **Header title dropped** on tiny to reclaim vertical space → taller 96 px panels.
- **Compact `5h` / `7d` pills** (were `Current` / `Weekly`) so the label doesn't collide with the number at 240 px.
- **Plain-ASCII status line** on tiny (`anim_plain`): the small styrene font lacks the spinner and ellipsis glyphs, which otherwise render as tofu boxes on both sides.
[...]

PR #111, opened 2026-07-15
Multi-account cycling + accent colors, token auto-refresh, stale indicator & Telegram usage alerts

bluh-jcrubioa:
Builds on the **SpotPear ESP32-S3-LCD-1.54 board port (#108)** and adds multi-account support, several UX/reliability features, and a Telegram usage-alert system. All verified end-to-end on the SpotPear hardware via framebuffer screenshots.

### Commits
- `SpotPear ESP32-S3-LCD-1.54 (MUMA) board port + tiny 240×240 layout` — also in #108
- `Multi-account cycling, accent colors, token auto-refresh + stale indicator`
- `Add per-account usage threshold alerts (Telegram)`

## Multi-account
- Optional `accounts.json` (`[{label, dir, color}, …]`): the daemon polls each account and sends one payload per account tagged `lbl`/`id`/`n`/`col`. No file → single-account (unchanged).
- Firmware stores up to 4 accounts; **tap the screen to cycle** between them (no timed auto-cycle). Each account's alias is the top header, in its own **accent color**.

## Reliability
- **Token auto-refresh** — on a 401 the daemon exchanges the stored refresh token for a fresh one (rotates + writes it back atomically) and retries, so an expired token self-heals instead of getting stuck.
- **Stale indicator** — dims the last numbers and shows an amber "stale" line when no fresh data has landed for >2.5 min.
- **BLE stability (board-scoped)** — 30s supervision timeout (`-DBLE_SUPERVISION_TMO`) + skip the optional refresh-subscription notify in multi-account mode; together these stop a cheap CSR dongle from flapping (it was dropping the link on transient GATT hiccups).

## Usage alerts (Telegram)
- Fires the first time any account crosses **50 / 80 / 100 %** on the **5h** or **7d** window, re-arming once that window resets (no spam). Each alert names the account and includes the full 5h + 7d readout with reset times.
- Config in a local `alerts.json` (bot token + chat id + optional thresholds); off unless configured. Crossing state persisted so restarts don't re-fire.

## Compatibility
Existing AMOLED boards keep their exact prior look — the large/compact `compute_layout()` branches are unchanged and every new field defaults to single-account / no-header behaviour. The boot-screen and supervision-timeout changes are gated behind board-scoped build flags.

https://claude.ai/code/session_01GFEedUyQB8UVb2HUVgmAdB


HermannBjorgvin:
Did you try the new Daemon's CLAUDE_CONFIG_DIR support for multiple accounts? It scans for any ~/.claude* dir and prompts you if you any to track it's usage. How is your handling of multiple accounts different?

4 accounts seems arbitrary also.

The telegram alerts feel out of scope for this project sorry. Trying to keep it focused on the hardware device and I don't want to take on the maintenance of it going forward.

The stale indicator could do with a photograph of it in the pull request I don't have time to flash the firmware on every pull request to test UX changes myself.

BLE stability could be good if you could split that into it's own PR.

There's been a lot of discussion on other pull requests on token refreshing and my stance generally on that is that this project should not risk breaking people's setups and stepping into Claude's token handling is an action this project shouldn't take. When a token is expired we assume the last usage stats and when a user starts a session with an expired token Claude handles refreshing tokens.

Thanks for the contribution tho. Curious to hear about the account part how other people are handling juggling multiple accounts.

HermannBjorgvin:
This PR is doing a bit too much to be merged as a single PR I suggest splitting up each feature into it's own PR if you want to get this merged.

PR #117, opened 2026-07-23
Report battery percentage over BLE

YarosMallorca:
Implement battery level reporting over BLE, allowing connected hosts to display the actual battery percentage instead of a constant 100%.

image


HermannBjorgvin:
Thanks for this PR, this is an awesome addition!

PR #120, opened 2026-07-24
Fix touch not tracking the auto-rotating display

sememtac:
Fixes a pre-existing bug on the AMOLED-2.16: the display auto-rotates (IMU-driven CPU pixel rotation in `display.cpp`, since the CO5300 can't rotate in hardware) but the touch point does **not** — the raw coordinate is fed straight to LVGL, so taps land in the wrong place whenever the device is turned. Only the un-rotated orientation is correct.

### Fix
Mirror the display's rotation into LVGL with `lv_display_set_rotation()` (set from `imu_hal_rotation_quadrant()` each loop) and let LVGL rotate the touch point for hit-testing itself (`lv_indev` → `lv_display_rotate_point`). All that remains is one fixed, quadrant-independent touch→panel alignment in the new `touch_rotate.{cpp,h}`; LVGL applies the per-orientation rotation on top — no per-quadrant calibration table.

- Display rotation (`rotate_strip`) is unchanged, and LVGL's matrix rotation stays off, so there's no double-rotation.
- Boards that don't auto-rotate report quadrant 0 → identity, no behavior change.

### Verification
Tested on an AMOLED-2.16 with a temporary logger: tapping the same on-screen corner in **all four orientations** maps to the same LVGL coordinate (it scattered before the fix). Builds clean on `waveshare_amoled_216`.

### Diff
Small and self-contained — `firmware/src/touch_rotate.{cpp,h}` (new) plus a handful of lines in `main.cpp` (3 files, +64/-2).

### DIFF
--- a/firmware/src/main.cpp
+++ b/firmware/src/main.cpp
@@ -20,6 +20,7 @@
 #include "hal/power_hal.h"
 #include "hal/imu_hal.h"
 #include "hal/sound_hal.h"
+#include "touch_rotate.h"
 
 static UsageData usage = {};
 
@@ -36,6 +37,20 @@ static UsageData usage = {};
 #endif
 static uint16_t* buf1 = nullptr;
 static uint16_t* buf2 = nullptr;
+
+static lv_display_t* g_disp = nullptr;   // for lv_display_set_rotation on IMU quadrant change
+
+// Map the board's IMU rotation quadrant to the LVGL rotation that lines touch
+// input up with the CPU-rotated display. (See touch_rotate.h — the display
+// rotates but touch reports native, so we let LVGL rotate the input to match.)
+static lv_display_rotation_t quadrant_to_rotation(uint8_t q) {
+    switch (q) {
+        case 1:  return LV_DISPLAY_ROTATION_270;
+        case 2:  return LV_DISPLAY_ROTATION_180;
+        case 3:  return LV_DISPLAY_ROTATION_90;
+        default: return LV_DISPLAY_ROTATION_0;
+    }
+}
 
 static uint32_t my_tick(void) { return millis(); }
 
@@ -59,10 +74,16 @@ static void rounder_cb(lv_event_t* e) {
 //           panel is dark, so pets/sleeves can't wake it overnight and LVGL
 //           can't quietly toggle splash<->usage on a black panel.
 static void my_touch_cb(lv_indev_t* indev, lv_indev_data_t* data) {
-    uint16_t x, y;
+    uint16_t rx, ry;
     bool pressed;
-    touch_hal_read(&x, &y, &pressed);
+    touch_hal_read(&rx, &ry, &pressed);
     const bool raw_pressed = pressed;
+
+    // Align the raw touch to the panel frame; LVGL applies the per-orientation
+    // rotation (lv_display_set_rotation, set from the IMU quadrant in loop()), so
+    // taps land correctly whatever way the display is turned.
+    uint16_t x, y;
+    touch_to_panel(rx, ry, &x, &y);
 
     if (IDLE_WAKE_ON_TOUCH) {
         static bool touch_was = false;
@@ -215,6 +236,7 @@ void setup() {
     buf2 = (uint16_t*)heap_caps_malloc(W * BUF_LINES * 2, LV_BUF_CAPS);
 
     lv_display_t* disp = lv_display_create(W, H);
+    g_disp = disp;
     lv_display_set_color_format(disp, LV_COLOR_FORMAT_RGB565);
     lv_display_set_flush_cb(disp, my_flush_cb);
     lv_display_set_buffers(disp, buf1, buf2, W * BUF_LINES * 2,
@@ -292,6 +314,13 @@ void loop() {
     ble_tick();
     power_hal_tick();
     imu_hal_tick();
+    // Mirror the display auto-rotation into LVGL so it rotates touch input to
+    // match (see touch_rotate.h). Set before display_hal_tick invalidates.
+    {
+        static uint8_t last_q = 255;
+        uint8_t q = imu_hal_rotation_quadrant();
+        if (q != last_q) { lv_display_set_rotation(g_disp, quadrant_to_rotation(q)); last_q = q; }
+    }
     sound_hal_tick();
     splash_tick();
     // Rotation transition (blank + ramp) would fight the idle fade — skip
--- a/firmware/src/touch_rotate.cpp
+++ b/firmware/src/touch_rotate.cpp
@@ -1,0 +1,15 @@
+#include "touch_rotate.h"
+#include "hal/board_caps.h"
+
+// Fixed touch-controller -> physical-panel alignment (quadrant-independent).
+// The CST9220 is configured with swap+mirror (touch.cpp), which lands its output
+// a quarter turn off the panel; this single transform undoes that so the value we
+// hand LVGL is in the panel's own frame. LVGL's lv_display_rotate_point() then
+// applies the per-orientation rotation. (First-pass value; tune here if a specific
+// orientation is uniformly off — it's ONE transform for all four quadrants now.)
+void touch_to_panel(uint16_t raw_x, uint16_t raw_y, uint16_t* out_x, uint16_t* out_y) {
+    const int S = board_caps().width;   // square panel
+    *out_x = raw_y;
+    *out_y = S - 1 - raw_x;
+}
+
--- a/firmware/src/touch_rotate.h
+++ b/firmware/src/touch_rotate.h
@@ -1,0 +1,18 @@
+#pragma once
+#include <stdint.h>
+
+// Orientation-aware touch, LVGL-owned rotation model.
+//
+// The AMOLED-2.16 auto-rotates the DISPLAY by IMU quadrant (the CO5300 can't
+// rotate in hardware, so display.cpp CPU-rotates the flush via rotate_strip), but
+// the touch controller keeps reporting in its fixed native frame. Left alone, taps
+// land in the wrong place whenever the device is turned.
+//
+// The fix is to mirror the display rotation into LVGL via lv_display_set_rotation()
+// (see main.cpp) and let LVGL rotate the touch point for hit-testing on its own
+// (lv_indev -> lv_display_rotate_point). Then all that's left here is one fixed,
+// quadrant-independent alignment from the touch controller's output to the panel
+// frame — LVGL applies the per-orientation rotation on top. Boards that don't
+// auto-rotate report quadrant 0, so this stays an identity there.
+
+void touch_to_panel(uint16_t raw_x, uint16_t raw_y, uint16_t* out_x, uint16_t* out_y);


JesusValeraDev:
Cross-posting a check I ran on an S3 2.16 while reviewing #135, which depends on this PR ([comment there](https://github.com/HermannBjorgvin/Clawdmeter/issues/135#issuecomment-5170997425)).

**The math is right.** I had per-quadrant touch data measured on this board from earlier work (labelled taps in all four orientations); the correct raw→logical transform was `3 - quadrant`, with `q0=(x,y)`, `q1=(S-1-y,x)`, `q2=(S-1-x,S-1-y)`, `q3=(y,S-1-x)`. Your composition — fixed `touch_to_panel()` of `(x,y)→(y,S-1-x)` plus LVGL rotation 270/180/90 for quadrants 1/2/3 — reproduces that table exactly in all four quadrants (`R∘R = q2`, `R∘180 = q1`, `R∘q1 = identity`). So the "first-pass value" comment in `touch_rotate.cpp` can be promoted to a confirmed one on the 2.16.

**One thing I'd change before merge.** The PR description says boards that don't auto-rotate report quadrant 0 → identity, no behaviour change. That's true of the LVGL half, but `touch_to_panel()` lives in shared `firmware/src/touch_rotate.cpp` and `main.cpp` calls it unconditionally, so all five boards get the 2.16's CST9220 quarter-turn — and `S = board_caps().width` is used for the Y axis, which is wrong on the 368×448 and 410×502 portrait ports.

It's invisible today because every board's only target is "tap anywhere" (the same immunity this PR describes for the rotation bug itself), but #135 §7 invites people to enable a scrollable list on the 1.8 / 2.06 / 2.16 C6, and there it would surface as drags going sideways. Moving the transform into each board's `touch.cpp` per CLAUDE.md rule 10 would keep it 2.16-only.

Also worth noting: `platformio.ini` pins `lvgl@^9.2.0`, which resolves to 9.5.0 today. The behaviour this PR relies on still holds there — input is rotated before hit-testing, now via `lv_display_rotate_point()` from `indev_pointer_proc` (`lv_indev.c:743`) — and the render path's own rotation stays gated behind `LV_DRAW_TRANSFORM_USE_MATRIX` + `lv_display_set_matrix_rotation()`, both off in this build, so there's still no double-rotation.

JesusValeraDev:
Here's the board-scoping change as a patch on top of `377bcbc`, if it's useful — apply, ignore, or take just the idea.

The alignment moves out of shared `touch_rotate.cpp` into the 2.16's own `touch_hal_read()`, `touch_rotate.{h,cpp}` go away, and `main.cpp` goes back to calling `touch_hal_read(&x, &y, &pressed)` directly. The `lv_display_set_rotation()` mirroring stays where you put it — that part genuinely is board-agnostic, since every non-rotating board reports quadrant 0 and gets the identity.

Three notes:

- **Behaviour on the 2.16 is unchanged.** Same transform, same position relative to LVGL's rotation — it just runs one function earlier, inside the HAL instead of after it. `board_caps().width` became `LCD_HEIGHT` (both 480 there, but it's the right bound for a Y coordinate).
- **The latch stays raw.** The transform applies to the output copies on each read rather than to the stored `touch_x/touch_y`, so re-reading a stale latch on later loops can't accumulate rotations.
- The rotation model your header explained moved onto `quadrant_to_rotation()`, and the CST9220-specific rationale now sits next to the code it describes.

Builds clean on all six environments on your branch:

| Env | |
| --- | --- |
| `waveshare_amoled_216` | SUCCESS |
| `waveshare_amoled_18` | SUCCESS |
| `waveshare_amoled_18_c6` | SUCCESS |
| `waveshare_amoled_216_c6` | SUCCESS |
| `waveshare_amoled_206` | SUCCESS |
| `waveshare_lcd_154` | SUCCESS |

`waveshare_lcd_154` is the one that makes the case best — a 240×240 port that the shared transform would also have quarter-turned.

<details>
<summary>Patch (<code>git apply</code> on top of 377bcbc)</summary>

```diff
diff --git a/firmware/src/boards/waveshare_amoled_216/touch.cpp b/firmware/src/boards/waveshare_amoled_216/touch.cpp
index 2538536..e66de5c 100644
--- a/firmware/src/boards/waveshare_amoled_216/touch.cpp
+++ b/firmware/src/boards/waveshare_amoled_216/touch.cpp
@@ -29,6 +29,23 @@ void touch_hal_init(void) {
     Serial.println("Touch init OK");
 }
 
+// Fixed touch-controller -> panel-frame alignment, quadrant-independent.
+// The CST9220 is configured with swap+mirror above, which lands its output a
+// quarter turn off the panel; this undoes that so what the HAL reports is in
+// the panel's own frame. LVGL then applies the per-orientation rotation on top
+// (lv_display_set_rotation, driven by the IMU quadrant in main.cpp), so the two
+// compose to the transform measured on hardware for each quadrant.
+//
+// This is CST9220 calibration, so it belongs to this board and not to shared
+// code — the other ports report panel-aligned coordinates already, and a
+// quarter turn applied to all of them would misplace every positional touch
+// target (and, on the portrait panels, mix up the two axis lengths).
+static inline void align_to_panel(uint16_t* x, uint16_t* y) {
+    const uint16_t rx = *x, ry = *y;
+    *x = ry;
+    *y = (uint16_t)(LCD_HEIGHT - 1 - rx);
+}
+
 void touch_hal_read(uint16_t* x, uint16_t* y, bool* pressed) {
     if (touch_data_ready) {
         touch_data_ready = false;
@@ -45,4 +62,5 @@ void touch_hal_read(uint16_t* x, uint16_t* y, bool* pressed) {
     *x = touch_x;
     *y = touch_y;
     *pressed = touch_pressed;
+    align_to_panel(x, y);
 }
diff --git a/firmware/src/main.cpp b/firmware/src/main.cpp
index 98ae990..5c35a05 100644
--- a/firmware/src/main.cpp
+++ b/firmware/src/main.cpp
@@ -20,7 +20,6 @@
 #include "hal/power_hal.h"
 #include "hal/imu_hal.h"
 #include "hal/sound_hal.h"
-#include "touch_rotate.h"
 
 static UsageData usage = {};
 
@@ -41,8 +40,17 @@ static uint16_t* buf2 = nullptr;
 static lv_display_t* g_disp = nullptr;   // for lv_display_set_rotation on IMU quadrant change
 
 // Map the board's IMU rotation quadrant to the LVGL rotation that lines touch
-// input up with the CPU-rotated display. (See touch_rotate.h — the display
-// rotates but touch reports native, so we let LVGL rotate the input to match.)
+// input up with the display.
+//
+// Boards that auto-rotate do it by CPU pixel remap in the flush path (the
+// CO5300 can't rotate in hardware), while their touch controller keeps
+// reporting in the panel's fixed frame. Mirroring the quadrant into LVGL makes
+// lv_indev rotate the incoming point before hit-testing and scroll-direction
+// handling, so taps and drags both follow the picture. The rotation is the
+// inverse of the display remap, hence 1 -> 270 and 3 -> 90.
+//
+// Any controller-specific alignment belongs in the board's touch HAL, not
+// here; boards that don't auto-rotate report quadrant 0 and get the identity.
 static lv_display_rotation_t quadrant_to_rotation(uint8_t q) {
     switch (q) {
         case 1:  return LV_DISPLAY_ROTATION_270;
@@ -74,17 +82,15 @@ static void rounder_cb(lv_event_t* e) {
 //           panel is dark, so pets/sleeves can't wake it overnight and LVGL
 //           can't quietly toggle splash<->usage on a black panel.
 static void my_touch_cb(lv_indev_t* indev, lv_indev_data_t* data) {
-    uint16_t rx, ry;
+    uint16_t x, y;
     bool pressed;
-    touch_hal_read(&rx, &ry, &pressed);
+    // The board's touch HAL reports in the panel's own frame; LVGL applies the
+    // per-orientation rotation on top (lv_display_set_rotation, set from the IMU
+    // quadrant in loop()), so taps land correctly whatever way the display is
+    // turned.
+    touch_hal_read(&x, &y, &pressed);
     const bool raw_pressed = pressed;
 
-    // Align the raw touch to the panel frame; LVGL applies the per-orientation
-    // rotation (lv_display_set_rotation, set from the IMU quadrant in loop()), so
-    // taps land correctly whatever way the display is turned.
-    uint16_t x, y;
-    touch_to_panel(rx, ry, &x, &y);
-
     if (IDLE_WAKE_ON_TOUCH) {
         static bool touch_was = false;
         static bool touch_wake_swallowed = false;
@@ -315,7 +321,7 @@ void loop() {
     power_hal_tick();
     imu_hal_tick();
     // Mirror the display auto-rotation into LVGL so it rotates touch input to
-    // match (see touch_rotate.h). Set before display_hal_tick invalidates.
+    // match (see quadrant_to_rotation). Set before display_hal_tick invalidates.
     {
         static uint8_t last_q = 255;
         uint8_t q = imu_hal_rotation_quadrant();
diff --git a/firmware/src/touch_rotate.cpp b/firmware/src/touch_rotate.cpp
deleted file mode 100644
index f631a17..0000000
--- a/firmware/src/touch_rotate.cpp
+++ /dev/null
@@ -1,15 +0,0 @@
-#include "touch_rotate.h"
-#include "hal/board_caps.h"
-
-// Fixed touch-controller -> physical-panel alignment (quadrant-independent).
-// The CST9220 is configured with swap+mirror (touch.cpp), which lands its output
-// a quarter turn off the panel; this single transform undoes that so the value we
-// hand LVGL is in the panel's own frame. LVGL's lv_display_rotate_point() then
-// applies the per-orientation rotation. (First-pass value; tune here if a specific
-// orientation is uniformly off — it's ONE transform for all four quadrants now.)
-void touch_to_panel(uint16_t raw_x, uint16_t raw_y, uint16_t* out_x, uint16_t* out_y) {
-    const int S = board_caps().width;   // square panel
-    *out_x = raw_y;
-    *out_y = S - 1 - raw_x;
-}
-
diff --git a/firmware/src/touch_rotate.h b/firmware/src/touch_rotate.h
deleted file mode 100644
index 2086db5..0000000
--- a/firmware/src/touch_rotate.h
+++ /dev/null
@@ -1,18 +0,0 @@
-#pragma once
-#include <stdint.h>
-
-// Orientation-aware touch, LVGL-owned rotation model.
-//
-// The AMOLED-2.16 auto-rotates the DISPLAY by IMU quadrant (the CO5300 can't
-// rotate in hardware, so display.cpp CPU-rotates the flush via rotate_strip), but
-// the touch controller keeps reporting in its fixed native frame. Left alone, taps
-// land in the wrong place whenever the device is turned.
-//
-// The fix is to mirror the display rotation into LVGL via lv_display_set_rotation()
-// (see main.cpp) and let LVGL rotate the touch point for hit-testing on its own
-// (lv_indev -> lv_display_rotate_point). Then all that's left here is one fixed,
-// quadrant-independent alignment from the touch controller's output to the panel
-// frame — LVGL applies the per-orientation rotation on top. Boards that don't
-// auto-rotate report quadrant 0, so this stays an identity there.
-
-void touch_to_panel(uint16_t raw_x, uint16_t raw_y, uint16_t* out_x, uint16_t* out_y);
```

</details>

PR #122, opened 2026-07-24
Make the costume generator easy to author new props

sememtac:
Improves the costume-authoring workflow (`tools/gen_costumes.py`): a `@costume("name")` decorator registers a prop in one place, named anatomy landmarks (`A.EYE_L`/`HEAD_TOP`/...) replace guessed pixel coords, per-prop docstrings, a "how to add a costume" walkthrough, and a generated `costumes_guide.png` (grid + labeled landmarks) to help place new props.

The generated `costumes.h` is **byte-identical** — no art or firmware change, purely a friendlier generator.

_Stacked on #121 (the costume feature) — review the top commit._

PR #124, opened 2026-07-24
Persist 24h usage history to flash across reboots

sememtac:
Widens the `usage_history` ring to 24h (1440 samples) and persists it to NVS (~3KB blob) every ~10 min, restoring it on boot so the Trend/Burn pages show real history across reboots instead of starting empty. Restored data is discarded if the daemon's wall-clock shows the newest sample is stale (>26h). `usage_history_add()` gains an epoch arg for the staleness check.

_Stacked on #122 → #121 — review the top commit._


HermannBjorgvin:
Weak title and description, temporally linked to your session which I have 0 context of. Unclear what if any human QA was done on this.

Example: Casually mentions trend/burn pages as if they are known existing features.

Commit history mentions mascot costumes, no mention of them in PR title and description.

Very sloppy PR sorry. This I think is a common problem when using agent is they are not aware enough of how to present their work to another person not privy to the current session.

PR #127, opened 2026-07-24
Add a situationally-aware talking mascot

sememtac:
Poke the corner Claude for a quip drawn from the live stats (session/weekly %, burn rate, reset countdown, time of day, costume, connection), with generic gags as rare treats. He also speaks up on his own on a jittered timer — a **Clawd chatter** setting on the System page picks Quiet / Balanced / Chatty (persisted).

Built as a self-contained, theme-agnostic module (`clawd_speech.{cpp,h}`): the UI snapshots the live stats into a `ClawdContext` and calls it. `clawd_speak()` doubles as a general feedback vector (e.g. announcing a session refill). The bubble dwell scales with length, wraps + grows for long lines, and finishes across page changes. The quip list is one documented function to edit.

_Stacked on #126 → #124 → #122 → #121. Builds on the 2.16._


HermannBjorgvin:
I don't have time to flash and test this. If you want to pitch a feature like this the best thing to do is take pictures or video of the feature to demo it.

Issue #97, opened 2026-07-01
Windows daemon: multi config-dir + active-plan parity (follow-up to #95)


HermannBjorgvin:
🤖 AI Generated:

Follow-up to #95, which added optional multi Claude config-dir polling with active-plan selection to the Linux (bash) and macOS/Linux (Python) daemons. The **Windows daemon** (`daemon/claude_usage_daemon_windows.py`) does not yet support it.

### Current state
- The Windows daemon already reads the config file (`%LOCALAPPDATA%\Clawdmeter\config`) for `clock` and `chime`, so the config-file plumbing exists.
- Token discovery is single-plan: `_windows_credential_candidates()` resolves ONE token from `CLAUDE_CREDENTIALS_PATH` / `CLAUDE_CONFIG_DIR` env overrides or a fixed fallback list.

### What to add (mirror #95)
- Read a `config_dirs` key (comma-separated list) from the config file; default to the existing single-token behavior when unset (no change for current users).
- `read_token_for(dir)` — resolve a token per configured dir (each dir's `.credentials.json`); keep the current candidate-list behavior for the default/unset case.
- A `PlanSelector` equivalent: "active" = the plan whose 5h session % rose most recently (monotonic poll counter, so the choice is sticky; a window reset/drop isn't activity; startup falls back to the highest current %).
- Poll each configured dir per cycle and send the active plan's payload.

### Tests
The Windows daemon has a solid pytest suite (`daemon/tests/test_windows_*.py`). Add selector + dir-parsing + token-per-dir tests, mirroring `daemon/tests/test_macos_multidir.py` from #95.

### Notes
- Windows config path differs from Linux/macOS (`%LOCALAPPDATA%\Clawdmeter\config` vs `~/.config/claude-usage-monitor/config`).
- Installer prompts are tracked separately (see the `install-windows.ps1` follow-up).

Issue #42, opened 2026-05-27
Anthropic changed their header names


paulroberson:
Anthropic changed their header names — the 5h/7d versioned headers the daemon expects don't exist anymore. The new headers are just unified without the time window infix. You need to patch poll_api in the daemon.

Edit the payload block (lines ~192-198):

claude_usage_daemon.py

Find this block:
python    payload = {
        "s": pct(hdr("anthropic-ratelimit-unified-5h-utilization")),
        "sr": reset_minutes(hdr("anthropic-ratelimit-unified-5h-reset")),
        "w": pct(hdr("anthropic-ratelimit-unified-7d-utilization")),
        "wr": reset_minutes(hdr("anthropic-ratelimit-unified-7d-reset")),
        "st": hdr("anthropic-ratelimit-unified-5h-status", "unknown"),
        "ok": True,
    }

Replace with:
python    payload = {
        "s": pct(hdr("anthropic-ratelimit-unified-overage-utilization")),
        "sr": reset_minutes(hdr("anthropic-ratelimit-unified-overage-reset")),
        "w": pct(hdr("anthropic-ratelimit-unified-fallback-percentage")),
        "wr": reset_minutes(hdr("anthropic-ratelimit-unified-reset")),
        "st": hdr("anthropic-ratelimit-unified-status", "unknown"),
        "ok": True,
    }

Then restart the daemon:
bashlaunchctl unload ~/Library/LaunchAgents/com.user.claude-usage-daemon.plist
launchctl load -w ~/Library/LaunchAgents/com.user.claude-usage-daemon.plist
tail -F ~/Library/Logs/claude-usage-daemon.out.log

HermannBjorgvin:
Have a look at #59 I think this is being discussed there, when you go into the extra usage limits the request headers change.

Intentional on anthropics behalf but also something we're looking at improving

PR #128, opened 2026-07-24
Show per-model weekly usage (Fable/Opus row) via the token-free OAuth usage endpoint

jkbdco:
## What

Claude Code's `/usage` view shows a per-model weekly limit alongside the all-models one (e.g. "Current week (Fable)"). This PR brings that to the device, and switches the Python daemons to a token-free data source:

- The **weekly panel** gains a slim per-model row — model name + bar + percent — between the main weekly bar and the reset line. The name comes from the API (`Fable`, `Opus`, …), so it adapts to whatever plan the account has. When no scoped bucket is reported (old daemon, or the unchanged bash daemon), the single-bar layout is exactly as before.
- The **macOS and Windows daemons** poll `GET /api/oauth/usage` — the endpoint the Claude Code client itself reads for `/usage`. It returns the same 5h/7d buckets as the rate-limit headers, plus the per-model scoped weekly limits, and costs zero tokens (the message-call approach burns ~1,440 tokens/day at the 60s cadence).

New payload fields (absent = firmware hides the row): `"m"` = scoped weekly percent, `"mn"` = model display name (truncated to 15 chars for the firmware's buffer).

## Layout rationale

A third full panel would overflow every breakpoint, so the model row lives inside the weekly panel. In split mode the weekly panel grows slightly taller (`split_panel_h`, per breakpoint) rather than squeezing its contents: this keeps the main weekly bar at the **same y offset as the session panel's bar**, so the two panels' repeated structure (pill → big % → bar → reset line) stays on one visual grid. Every breakpoint has clearance below the weekly panel for the growth, and the height is restored whenever split mode ends.

## The endpoint history — called out explicitly

`/api/oauth/usage` was adopted once before (#29) and reverted (#37) after a rate-limit report. That incident involved a 5s retry loop that re-hammered the endpoint on every failure. This PR's approach differs in three ways:

1. **Fixed 60s cadence** — the endpoint is hit at most once per poll cycle per config dir, never in a retry loop.
2. **429 cooldown** — a rate-limit response benches the endpoint for 15 minutes (`USAGE_ENDPOINT_COOLDOWN_S`) while the fallback carries polling, with a regression test asserting no network calls happen during cooldown.
3. **Automatic fallback** — any failure or unrecognized shape falls back to the existing `/v1/messages` header method that cycle, so behavior can never be worse than today.

The endpoint is still undocumented and could change shape; the fallback is the safety net for that too (related: #42 — the headers themselves have been renamed before).

Two deliberate contract decisions, since this endpoint is unproven for edge cases:

- **Enterprise accounts**: the endpoint is only trusted when the response has **both** `five_hour` and `seven_day` utilization (Enterprise has no weekly concept). Anything else routes to `poll_api()`, which remains the sole authority on Enterprise detection.
- **Windows toast**: `poll_usage_endpoint()` never raises `AuthError` — on 401/403 it returns `None` and falls back, so a genuinely dead token still fires the "token expired" toast from `poll_api()`'s proven path, and an endpoint-specific rejection can never fire a false toast.

The Linux bash daemon is unchanged (still header-based, no per-model row) — parsing nested JSON in bash without a jq dependency felt out of scope. Happy to file a parity follow-up issue like #97/#98.

## Testing

- Live on a Max account: endpoint polled every 60s across an extended session; values match the header method; ~0.3s round trip.
- **Hardware-verified on all three layout breakpoints**, all with identical canned data for apples-to-apples comparison:
  - **Large (480×480)** — ESP32-C6-Touch-AMOLED-2.16, live daemon data:

    Large layout on AMOLED-2.16
  - **Compact (368×448)** — ESP32-S3-Touch-AMOLED-1.8, on-device framebuffer capture via the `screenshot` serial command:

    Compact layout on AMOLED-1.8
  - **Small (240×240)** — rendered via a temporary geometry override on the same S3 hardware and captured from the framebuffer (no 1.54 unit on hand; the capture exercises the exact layout code path):

    Small layout
- **Old-daemon compatibility on hardware**: an unmodified header-method daemon pushing to the new firmware renders the unchanged single-bar weekly (verified incidentally when one connected mid-test).
- **Unit tests**: `daemon/tests/test_usage_endpoint.py`, parametrized over both Python daemons (36 cases: nominal + scoped-model parsing, name truncation vs the firmware buffer, fallback signaling for every non-Pro/Max outcome, 429 cooldown incl. no-network-during-cooldown, malformed `resets_at` degradation per the #104/#106 guard precedent). Full daemon suite passes.
- All seven firmware envs build; three regression-built on this branch.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,6 +170,8 @@ See `~/.claude/projects/.../memory/` files for persistent context (user is an em
 
 Bash daemon (`daemon/claude-usage-daemon.sh`) reads OAuth token, polls Anthropic API, sends JSON over BLE GATT. Run with `systemctl --user start claude-usage-daemon`. The unit file's `ExecStart` is the absolute path to the script — repoint it when switching between the worktree and the main checkout.
 
+**Polling paths (macOS/Windows Python daemons):** `poll_usage_endpoint()` first — GET `/api/oauth/usage` (token-free; same buckets as the headers plus per-model `weekly_scoped` limits, sent as payload `"m"`/`"mn"`). Undocumented endpoint, so any failure or non-Pro/Max shape falls back to `poll_api()` (1-token `/v1/messages`, rate-limit headers) — which also remains the sole path that detects Enterprise accounts and (on Windows) the sole raiser of `AuthError` for the token-expired toast. A 429 from the usage endpoint benches it for 15 min (`USAGE_ENDPOINT_COOLDOWN_S`) — it was reverted once before over rate limiting (PRs #29/#37); don't remove the cooldown. The bash daemon still uses only the header method (no `m`/`mn`).
+
 **Discovery & resilience:**
 
 - Connects by name (`"Clawdmeter"`) on first run, caches resolved MAC at `~/.config/claude-usage-monitor/ble-address`. ESP32 BLE addresses are factory-burned per-chip, so swapping any board invalidates the cache.
--- a/README.md
+++ b/README.md
@@ -189,8 +189,7 @@ reg delete "HKCU\Software\Microsoft\Windows\CurrentVersion\Run" /v Clawdmeter /f
 ## How it works
 
 1. The daemon reads your Claude Code OAuth token — from the macOS Keychain (service `Claude Code-credentials`) on macOS, or from `~/.claude/.credentials.json` on Linux (`%USERPROFILE%\.claude\.credentials.json` on Windows).
-2. It makes a minimal API call to `api.anthropic.com/v1/messages` — one token of Haiku, basically free.
-3. The usage numbers come straight out of the response headers (`anthropic-ratelimit-unified-5h-utilization` and friends).
+2. It polls `api.anthropic.com/api/oauth/usage` (macOS/Windows daemons) — the same endpoint the Claude Code client's `/usage` view reads. It costs zero tokens and also carries per-model weekly limits (the "Fable"/"Opus" row on the display). Heads-up: this endpoint isn't publicly documented, so it could change; if the call fails for any reason the daemon automatically falls back to the original method — a minimal call to `api.anthropic.com/v1/messages` (one token of Haiku, basically free) whose response headers (`anthropic-ratelimit-unified-5h-utilization` and friends) carry the same session/weekly numbers, minus the per-model bucket. The Linux bash daemon still uses the header method.
 4. The daemon connects to the ESP32 over BLE and writes a JSON payload to the GATT RX characteristic.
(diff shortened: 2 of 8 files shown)


HermannBjorgvin:
The gaps on the new design look cramped. Maybe instead of fitting everything into one screen just add another view you can switch to by tapping the screen which could maybe contain the kind of usage you are wanting to monitor

HermannBjorgvin:
It has to look purposeful, not just like an afterthought added to the current design.

jkbdco:
I noticed that too after creating the PR so flipped it to Draft while fixiing it. Also tested it on a 1.8. Will mark it as ready for review shortly. Thanks!

jkbdco:
@HermannBjorgvin bumping this — been running it daily since I opened the PR, ~4000+ updates over the week and the Fable row worked the whole time. Funny enough my oauth token expired while I was away for a few days and the endpoint started 429ing, and the cooldown + fallback just did their thing — no hammering, recovered on its own when I logged back in. So the thing that got #29 reverted actually happened to me and the guardrails held.

On the separate view idea — after a week of using it I'd push back a little. The Fable weekly is always the first limit I hit, so it's the number I want at a glance. Putting it behind a tap hides the thing I check most. Happy to rework it that way if you still prefer it though. Thanks!

HermannBjorgvin:
@jkbdco agreed on the fable limit, it's the one I'm watching most closely these days.

I opened #136 due to a lot of feedback from people wanting to see more detailed per session overviews. It expands the session overview to showing the model behind a session and the context window.

Maybe this could be incorporated into that, i.e. if a session is running on fable add a fable usage bar, I think that would be useful to have.

But that might be a pain to do there since readability is a bit of an issue. anyway sorry for getting sidetracked there.

What about this since I didn't really like the smaller fable loading bar what about reusing the weekly loading bar of the main screen and showing a terracotta colored loading bar in the same place as the weekly loading bar, so it would just share the space with it as an overlay and then we can add a terracotta colored *fable asterisk aligned right in the same line as the weekly "Resets in..." text?

I think that would make it seem like less of an afterthought than adding the additional loading bar below the weekly one

HermannBjorgvin:
Just realized the coloring for the weekly bar would conflicts i.e. going from green/terracotta/red based on the percentage used weekly but perhaps we can just remove the warning colors on the weekly since it is not as relevant there, if you have 80% used of your weekly but only one day left to reset it doesn't really warrant a warning color.

HermannBjorgvin:
Closing, on the 14th of September the Fable model will no longer have a separate limit. We can reopen if the policy changes back in the future.

PR #133, opened 2026-07-30
feat: replace screenshot.sh with a cross-platform screenshot.py

Fr3d:
## Why

`screenshot.sh` needs bash plus `ffmpeg`, so it can't be run from PowerShell — which makes the documented "screenshot every UI iteration" workflow unavailable on Windows.

## What

A Python rewrite with the same positional arguments, so existing references still read the same:

```
py screenshot.py                        # auto-detect port
py screenshot.py sessions.png           # chosen filename
py screenshot.py sessions.png COM4      # explicit port, as the .sh took
py screenshot.py --list                 # show candidate ports
```

- **No ffmpeg, no Pillow.** PNG is written with stdlib `zlib`/`struct`. `pyserial` is the only dependency, and the script re-execs under PlatformIO's bundled Python when it's missing — the same fallback the shell script used.
- **Port auto-detection** by Espressif USB VID `0x303A`, so it finds the board without being told and won't grab an unrelated COM port.
- **Clearer failures**: `SCREENSHOT_UNSUPPORTED` now explains itself (C6 boards have no PSRAM for a framebuffer) instead of surfacing as a timeout.

Deletes `screenshot.sh`, since the Python version supersedes it on all three platforms. Adds a `.gitignore` rule for root-level PNGs (capture output); `screenshots/` stays tracked.

## Two details worth reviewing

**DTR/RTS are held low before opening the port.** On the ESP32-S3's native USB-JTAG, toggling them resets the chip — which reboots the device to the splash and destroys the screen you were trying to capture. The bash version never hit this because pyserial's defaults happen to be harmless on the Linux/macOS hardware it ran on.

**RGB565 channels expand by bit replication** (`(v << 3) | (v >> 2)`) rather than a plain shift, so panel white lands on `#FFFFFF` instead of `#F8F8F8`. That matters when checking text colours against the theme.

## Verified

Captured a real 480×480 frame from an AMOLED-2.16 over COM4 on Windows.

Not verified on macOS or Linux — I don't have either to hand. The port-detection fallbacks for `usbmodem`/`ttyACM` are written but untested, so a second pair of eyes there would be welcome.


HermannBjorgvin:
You can't just use WSL?

Fr3d:
WSL2 can't access COM ports natively. Apparently there are some hacky workarounds, most of which need admin elevation to achieve. None of them are recommended.

Officially WSL1 does support COM ports, though there are open bugs for it them not working. Additionally WSL1 is no longer the default when you install WSL, and I'd assume it's likely to be deprecated by MS soon.

Finally, switching between WSL1 and WSL2 is not quick or painless for the user.

PR #134, opened 2026-07-30
feat: add a Sessions screen showing running Claude Code sessions

Fr3d:
## What it does

A third screen alongside the splash and usage views, showing every running Claude Code session and what it's doing — thinking, running a tool, needing permission, asking you a question, or dead on a rate limit — with its context-window usage.

Sessions

Reached by touch: usage → sessions → splash. Touch was already the screen-switching gesture, so the new screen joins that cycle rather than the PWR button, which keeps its existing jobs (brightness, next animation) — no existing gesture changes meaning.

**Off by default.** Set `hook_port` in the daemon config and add a hook block to `~/.claude/settings.json`; both are documented in the README.

## How it gets the data

Claude Code can POST hook events to an HTTP endpoint (`"type": "http"` hooks). The daemon listens on loopback and maintains a session table, then writes it to a new GATT characteristic.

Hooks rather than watching the transcript files, for two reasons. A session blocked on a permission prompt writes **nothing** to its transcript until you answer, so that state is invisible to any file-watching approach — the `Notification` event reports it directly. And hooks arrive in milliseconds rather than on a poll interval.

Liveness and naming come from `~/.claude/sessions/<pid>.json` rather than being inferred. That file exists only while a process runs, so a vanished session retires within seconds whatever the cause — a timeout alone was wrong in both directions, evicting sessions left idle overnight while they were alive, and letting crashed ones linger for hours showing a stale "idle". It also supplies Claude Code's own per-session name, which is unique where `basename(cwd)` is not: two sessions in one directory read as `clawdmeter-36` and `clawdmeter-2c` instead of `clawdmeter` twice.

## Design decisions you may want to push back on

**Colour is rationed.** Only the four states that want you get terra-cotta; active work is plain text colour and idle is dim. The screen's job is answering "does anything need me" without being read, so more colour than that defeats it. This also meant overriding `make_bar()`'s green indicator, which is the usage screen's headroom language and competed for attention here.

**Sessions get their own characteristic** (`…0005`) rather than sharing the usage write. That payload already reaches 121 bytes worst case, which would leave only ~79 bytes of the MTU budget for rows — about two.

**The row budget is measured, not assumed.** `budget_from_mtu()` derives it from `client.mtu_size` per connection. Windows/WinRT negotiates 256 (253 usable), which affords 20-character labels; CoreBluetooth and BlueZ are unmeasured, so each falls back to a conservative default rather than trusting the Windows figure. `daemon/probe_mtu.py` is included for measuring.

**Labels shrink before rows are dropped.** Hiding a session is the one thing this screen must not do, so a tight budget costs label characters first. Truncation middle-elides to preserve Claude Code's trailing discriminator — tail-truncating `somelongprojectname-2c` would put the collision straight back.

**Context % is a heuristic and I'd like a second opinion.** Tokens come from the newest non-sidechain assistant record's `input_tokens + cache_read + cache_creation`, but transcripts record `message.model` with any `[1m]` suffix stripped, so window size can't be read from the name — a 1M session displayed 100% until the token count itself was used as evidence of the limit. The authoritative figure is the statusline's `context_window.used_percentage`, which hooks don't provide.

## Security

The listener binds `127.0.0.1` only and rejects non-loopback peers, because hook payloads contain prompt and response text. The subscribed hooks are read-only observers: the endpoint answers `204 No Content`, so they cannot block a tool call or auto-approve a permission. On the device side, the new characteristic goes through the same single-owner check as the usage write rather than becoming an unguarded back door.

## Platform coverage

- **Windows** — verified end to end on an AMOLED-2.16.
- **macOS** — code paths present, untested.
- **Linux** — the bash daemon has no listener path, so Linux users would run `claude_usage_daemon.py` instead. That should work (its macOS-specific pieces are all behind platform guards or lazy imports, and Linux credential/address handling is already deliberate), but note `discover_target()` never scans by name off darwin, so it needs an address already pinned in `~/.config/claude-usage-monitor/ble-address` and can't populate that itself. Untested against BlueZ. Adding a file-drop bridge to the bash daemon would fit its existing dbus-monitor→flag-file idiom, and I'm happy to do that if you'd prefer it in scope.

## Verified

238 tests pass; firmware builds for `waveshare_amoled_216` (RAM 31.9%, flash 54.8%). Screenshot above is from real hardware.

**Not verified:** the other four board ports. The changes are mostly in shared code, and the two C6 boards have no PSRAM, so the five pre-built row groups land in internal RAM there — worth a look from someone with that hardware. A 20-character label has also never been rendered, so whether it collides with the context bar at any breakpoint is untested.


HermannBjorgvin:
Take a look at #129 and the feedback I had on that PR. This is tackling the same topic but with useful additional info like the user prompt feedback which looks particularly useful.

HermannBjorgvin:
Closing in favor of #136 which consolidates this and a couple of other pull requests and feature requests I've gotten on this topic. Would appreciate any help on testing that one since it's a large change so any help validating it would be appreciated.

Issue #135, opened 2026-07-31
Live session awareness and multiple session overview


HermannBjorgvin:
# Live session awareness and multiple session overview

---

Today the device shows how much quota you've burned. This adds **what Claude
Code is doing right now**: which chats are open, how full each one's context
window is, how far through its todo list it is, how many subagents it has
running — and, most importantly, whether any of them is stuck waiting on you.

The design constraint that shapes everything below: **you never operate it.**
The display picks its own layout and its own ordering from Claude's state. There
is no menu, no mode switch, no paging between sessions, and no setting for which
view you prefer. When a chat needs you, it is already in the top slot with a
pulsing indicator, whether or not you were looking.

Two things to know before reading on:

**It's opt-in per board.** Gated behind a capability flag (§7) and enabled on
the 480×480 AMOLED-2.16 only to start with. It assumes a panel with room for
three chat cards and a refresh model that tolerates animation, which rules out
the small ports and would rule out an e-ink panel whatever its resolution. Boards
without the flag keep today's behaviour exactly — no dead UI, no wasted RAM.

**It's off by default even there**, because the session data comes from a Claude
Code hook integration you have to install. §9 covers what the device does in
each partial configuration — no hooks, no transcript access, no rotation fix, no
scrolling — since most of those are states a real user passes through rather
than error cases.

---

## 1. What you see

### 1.1 Anatomy of a chat row

Every chat is drawn the same way, in FOCUS and in CHATS. Three lines:

```
    ┌──────────────────────────────────────────┐
    │ ordora-platform                      25% │  ← name, context %
    │ ▓▓▓▓▓▓▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ │  ← context bar
    │ ● running Bash          ☑2/4  ⚇3     10s │  ← state line
    └──────────────────────────────────────────┘
      │ │                     │     │      │
      │ │                     │     │      └─ time in this state
      │ │                     │     └──────── subagents running
      │ │                     └────────────── todos done / total
      │ └──────────────────────────────────── what it's doing
      └────────────────────────────────────── state indicator
```

- **`●` state indicator** — dim when idle, neutral when working, terra-cotta and
  **pulsing** when the session needs a human.
- **`☑` todo badge** — [Lucide `list-todo`](https://lucide.dev/icons/list-todo)
  plus `done/total`, from the session's most recent `TodoWrite`. Hidden when the
  session has no todo list.
[...]

PR #136, opened 2026-08-01
Live session awareness and multiple session overview

HermannBjorgvin:
Implements #135.

The device now shows what Claude Code is doing right now: which chats are open,
how full each context window is, todo and subagent counts — and, above all,
whether any chat is stuck waiting on you. Nothing is operated by hand: the
display picks its own layout and ordering from Claude's state, and a chat that
needs you is always in the top slot with a pulsing indicator.

The feature is opt-in twice over: gated per board by a capability flag
(enabled on the 480×480 AMOLED-2.16 to start), and off by default on the
host until the hook integration is configured. Without either, the device
behaves exactly as it does on `main`.

## On-device views (firmware)

No new screen. The usage screen's resolver (`update_view_state()` in
`firmware/src/ui.cpp`) selects among five sub-views: pairing hint, no-data,
quota panels (RESTING), ONE CHAT, and SEVERAL CHATS, per the §2.1 flowchart in
the issue. A waiting chat pins the view; after the last chat closes the view
lingers 10 minutes (`CHAT_LINGER_MS`) before returning to RESTING.

- **Chat cards** (`firmware/src/ui.cpp`): name (ellipsized firmware-side
  against a worst-case `sonnet-5.2` model pill), context bar, and a state line
  — `●` indicator plus state text on the left, a right-aligned cluster of
  todo badge (terra-cotta), subagent badge (purple `#9e87be`, hue-matched to
  the palette), and time-in-state.
- **ONE CHAT**: the 5h quota panel as its own card, then the chat card with
  `95% of context used`, the absolute token count (`190K` / `1.2M`), and a
  model pill. The long-form wording deliberately teaches what the terse
  multi-chat bars mean.
- **SEVERAL CHATS**: a one-line 5h/7d quota strip, then up to 3½ cards at
  88px pitch — the clipped fourth card is the "more below" affordance. The
  host pre-sorts (waiting → working → idle); the firmware renders received
  order and never re-sorts.
- **Motion budget**: waiting cards pulse indicator + state text on one shared
  700ms phase; rank changes slide cards 260ms ease-out with stable
  sid-keyed identity; content refreshes move nothing. Idle cards dim to 60%
  opacity via plain `opa` (not `opa_layered`, which would allocate composite
  buffers).
- **Wake on activity**: every session payload resets the idle-fade timer
  (`idle_note_activity()` in `firmware/src/main.cpp`), so the panel is awake
  while Claude works and dark when nobody is.
- **Board gating**: `BOARD_HAS_SESSION_VIEWS` compile-time flag + runtime
  `board_caps().has_session_views` (documented in
  `docs/porting/capability-flags.md`). Flag-off boards compile the card pool
  out entirely — verified: the C6 build contains zero session symbols.
  Flag-on envs run `LV_MEM_SIZE=128K` (the ~90-widget card pool exhausts
  LVGL's 64K default pool).
- **Icons**: Lucide `list-todo` / `users-round` at 24/16px, RGB565A8 with a
  real alpha plane, recolored at render time so the 60% dim rides the normal
  draw path (`firmware/src/icons.h`, sources in `assets/`).

## BLE wire

Session rows arrive on a new SS characteristic
`4c41555a-4465-7669-6365-000000000005` (`firmware/src/ble.cpp`), behind the
same single-owner write guard as the existing RX characteristic, with
`setMTU(517)` and a 1KB receive buffer. Payload:

```jsonc
{"ss":[
  // [sid, label, state, ctx%, elapsed_s, model, tool, ntools, nagents, tdone, ttotal, tok]
  ["a3","billing-svc-3f",6,95,62,1,0,0,0,3,7,190]
]}
```

State codes 0–10 are append-only wire constants (`firmware/src/data.h`).
`ctx` −1 hides the bar rather than drawing it empty; `tok` is context tokens
in 1K units (−1 or absent → the card falls back to showing the percentage).
Malformed payloads are dropped and the last good list stays on screen
(`firmware/src/main.cpp`).

## Host: session sidecar (daemon)

`daemon/clawdmeter_sessions.py` — a stdlib-only Python hook listener, usable
as a library or standalone:

- Claude Code hooks POST to `127.0.0.1:<hook_port>`; the listener is a
  read-only observer (answers 204, rejects non-loopback peers, can never
  block or approve a tool call).
- State machine over the issue's §4.1 event table, plus `PostToolUseFailure`
  so a failed tool can't leave the concurrent-tool count stuck. Tool state
  clears on `Stop`/`UserPromptSubmit`, not `PostToolUse`, so the state line
  doesn't flicker between tools.
- Liveness from the session roster (`<config-dir>/sessions/<pid>.json`,
  checked against pid + process start time so pid reuse can't fake a live
  session), with a 30s grace and a 6h staleness sweep — never from activity
  timeouts, because a chat blocked on a permission prompt is silent.
- Context from the transcript's newest non-sidechain assistant record;
  window heuristic: 200k default, `[1m]` → 1M, snap-up on overflow,
  `context_window_k` config pins it. `tok` comes from the same read as the
  percentage so the two can never disagree.
- Sort `(bucket, -last_event_at)`; MTU fitting middle-elides labels to an
  8-char floor (trailing discriminators survive) before dropping rows from
  the tail, so a waiting chat is never the one dropped.
- Writes the fitted payload atomically to `~/.clawdmeter/sessions.json`; the
  bash daemon (`daemon/claude-usage-daemon.sh`) ships it to the SS
  characteristic on its existing 5s tick with a byte-exact UTF-8 GATT write.
  Fully inert when the sidecar isn't running or the firmware lacks the
  characteristic.
- `daemon/clawdmeter-sessions.service` (systemd user unit),
  `hook_port` / `context_window_k` / `sessions_budget_bytes` in
  `daemon/config.example`, an `install.sh` prompt that merges the hook block
  into `~/.claude/settings.json` via `--install-hooks`, and setup docs +
  wire-format reference in `daemon/SESSIONS.md`.

## Scope

- **No scrolling.** §2.5 marks it optional and it hard-depends on the touch
  rotation fix (#120); ordering already keeps anything urgent above the fold.
- **Hooks-only.** No transcript-fallback mode (§11.2); without `hook_port`
  the device is exactly today's device.
- **Boards**: enabled on `waveshare_amoled_216`. Other boards follow the §7
  checklist (flag flip + hardware look).

## Verification

- All views screenshot-verified against the firmware running in a desktop
  SDL2 build (attached), including pixel-level alignment checks and
  pulse-phase captures. **Not yet verified on 2.16 hardware** — needs the §7
  pass (three chats, permission block, reorder smoothness) on a real panel
  before merge.
- Daemon: 50 unit tests (`pytest daemon/tests/test_sessions.py`) covering the
  state machine, sort, eliding, MTU ladder, and window heuristic, plus a
  loopback smoke test driving the listener with curl'd hook sequences.

## Screenshots

| File | Shows |
| --- | --- |
|image| RESTING — today's quota view, unchanged |
|image| ONE CHAT — 5h panel + chat card, badges |
|image| Waiting: accent + pulse, status line hidden |
|image| Sorted list, waiting chat pinned to top, clipped 4th card |
|image| Working chats + 60%-dimmed idle card |
|image| The 260ms reorder slide mid-animation |
|image| Name ellipsis against the model pill |
|image| Million-scale token formatting |


HermannBjorgvin:
Don't have a board on me so this is untested on real hardware. Will test it on actual hardware on Tuesday when I'm back at work.

HermannBjorgvin:
Still haven't looked at how the daemon and hooks work on this PR. I think it's likely I'll have some opinions on those but going to look at it again a bit later

HermannBjorgvin:
Readability is a concern here, the UI might need to be reworked a bit so the user doesn't need a microscope to monitor the sessions.

Issue #98, opened 2026-07-01
install-windows.ps1: interactive config prompts (dirs/clock/chime) parity (follow-up to #96)


HermannBjorgvin:
🤖 AI Generated:

Follow-up to #96, which added interactive config prompts to `install.sh` (Linux) and `install-mac.sh` (macOS): pick which config dirs to poll, plus the optional clock and session-reset chime. **`install-windows.ps1`** has no equivalent — users hand-edit `%LOCALAPPDATA%\Clawdmeter\config`.

### What to add (mirror #96, in PowerShell)
- Detect `%USERPROFILE%\.claude*` config dirs that hold credentials and, when more than one exists, prompt which plans to poll → write `config_dirs`.
- Prompt for the optional `clock` (`off/auto/12/24`) and `chime` (`y/N`).
- Upsert keys into `%LOCALAPPDATA%\Clawdmeter\config`, preserving other keys and idempotent; pressing Enter through the defaults writes nothing.

### Dependencies / notes
- **`clock` and `chime` already work** on the Windows daemon, so those two prompts are immediately useful.
- **`config_dirs`** only takes effect once the Windows daemon supports multi config-dir polling — see #97. The prompt can be added regardless, but land/track #97 alongside it.
- Match the interactive/non-interactive guarding of the Linux/macOS installers (skip prompts when not attached to a console).

PR #140, opened 2026-08-03
feat(daemon): Windows multi config-dir polling with active-plan selection (#97)

### LINKED ISSUE #97: Windows daemon: multi config-dir + active-plan parity (follow-up to #95)

ChinesePrince07:
Implements #97 — mirrors #95's multi config-dir + active-plan selection into the Windows daemon.

**What it does**
- New `config_dirs` key (comma-separated) in `%LOCALAPPDATA%\Clawdmeter\config`, parsed with the same style as the existing `chime`/`clock` readers.
- `PlanSelector` ported verbatim from #95's Python daemon: active = plan whose 5h session % rose most recently (rise stamps a monotonic poll counter, so the choice is sticky; a window reset to 0 is not activity; startup/tie falls back to highest current %). Module-level instance so state survives reconnects, matching #95's `_SELECTOR`.

**Deliberate deviations from #95, called out for review**
1. Unset `config_dirs` returns `[]` and routes through the *untouched* legacy `read_token()` path — the Windows default is a candidate/probe LIST (env overrides + 3 paths), not a single dir, so substituting `[~/.claude]` would change current behavior. Zero change for existing setups; the pre-existing reconnect tests pass unmodified.
2. Toast semantics (no #95 analogue, macOS has no tray): "token expired" toasts only when NO configured dir yields a payload AND the failure is genuinely token-shaped (missing or 401/403 everywhere). All-transient cycles stay silent (preserves the SC#5 fix), and one expired sibling never hides a working plan.
3. No Keychain fallback in `read_token_for` — no Windows analogue.

**Testing**
- 26 new tests in `daemon/tests/test_windows_multidir.py` following the existing mock conventions: config parsing edge cases, selector rules (startup/rise/sticky/reset), legacy-path preservation, multi-dir polling, and all four toast-semantics cases.
- Full daemon suite: **142 passed**.
- Same caveat as #139: unit-tested (Linux/pytest), not run on a physical Windows machine — the poll/selection logic is exercised through the real code paths with mocked BLE/API/filesystem.

Not included: `install-windows.ps1` prompts (#98, tracked separately per the issue notes) and Windows config docs (no existing Windows doc surface documents any config keys yet — happy to add a section to `README-windows.md` in this PR if you want one).

Closes #97

### DIFF
--- a/daemon/claude_usage_daemon_windows.py
+++ b/daemon/claude_usage_daemon_windows.py
@@ -507,6 +507,142 @@ def _read_expiry() -> str:
     return "expiry unknown"
 
 
+# --- Multi config-dir support (#97, mirrors #95) ---------------------------
+# Claude Code can run against more than one config dir (e.g. ~/.claude for a
+# personal plan and ~/.claude-work for a work plan, selected via
+# CLAUDE_CONFIG_DIR). When the `config_dirs` option is set, the daemon polls
+# each dir's token every cycle and shows whichever plan is "active".
+
+
+def read_config_dirs() -> list[Path]:
+    """Claude config dirs to poll, from the `config_dirs` option (comma list).
+
+    Returns [] when unset/blank so the caller keeps the existing read_token()
+    candidate-list behavior (env overrides + probe list) — zero change for
+    current single-plan setups (#97). Unlike #95's Linux/macOS daemons there
+    is no single default dir to substitute here (the default is a probe LIST),
+    hence [] rather than [~/.claude]. ~ is expanded; blank entries dropped.
+    """
+    raw = ""
+    try:
+        if CONFIG_FILE.exists():
+            for line in CONFIG_FILE.read_text().splitlines():
+                line = line.split("#", 1)[0].strip()
+                if "=" not in line:
+                    continue
+                key, val = line.split("=", 1)
+                if key.strip().lower() == "config_dirs":
+                    raw = val.strip()
+    except OSError:
+        pass
+    if not raw:
+        return []
+    return [Path(p.strip()).expanduser() for p in raw.split(",") if p.strip()]
+
+
+def read_token_for(config_dir: Path) -> str | None:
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
Closing due to not having tested on real hardware. This project relies on testing on physical devices and real OS.

PR #141, opened 2026-08-03
docs: list the LCD-1.54 port in CLAUDE.md's board inventory

nathanjohnpayne:
`CLAUDE.md` is the bootstrap doc for agent sessions, but its board inventory stopped at five ports while `firmware/src/boards/` holds six — the ESP32-S3-Touch-LCD-1.54 port and its `[env:waveshare_lcd_154]` block were missing, so a session reading this file wouldn't know the board exists.

This adds the entry in all three places the doc enumerates boards (the port list, the architecture tree, and the build commands) and bumps the counts to **six ports / four panel sizes**.

The entry calls out what makes this port unlike the others, since those are the things that bite:

- **First non-AMOLED port** — a plain 4-wire SPI ST7789, not QSPI.
- **No panel brightness command** — backlight is LEDC PWM on `LCD_BL`.
- **No PMU** — battery is an ADC divider on GPIO1, and `BAT_EN` (GPIO2) is a power-hold line that must be driven HIGH early in `board_init()` or the board browns out on battery.

`README.md` already lists the board (#102); only `CLAUDE.md` was missed. Docs-only — no code changes. Verified the board still builds (`pio run -e waveshare_lcd_154` → SUCCESS) and every claim in the entry is checked against `board.h`, `caps.cpp`, `sound.cpp`, and the PlatformIO env.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,13 +6,14 @@ selected via PlatformIO's `build_src_filter`. Adding a board means dropping in
 a new folder + a new `[env:...]` block — `main.cpp`, `ui.cpp`, and `splash.cpp`
 never see board-specific code. See [`docs/porting/adding-a-board.md`](docs/porting/adding-a-board.md).
 
-Five ports today (two SoC families, three panel sizes):
+Six ports today (two SoC families, four panel sizes):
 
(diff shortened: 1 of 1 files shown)


HermannBjorgvin (approved):

HermannBjorgvin:
Thanks for the fix @nathanjohnpayne !

PR #143, opened 2026-08-03
Show the weekly Fable allowance on the usage screen when the plan has one

nathanjohnpayne:
Claude plans that include Fable now have a second weekly limit — the usage settings show two bars under Weekly limits, "All models" and "Fable", sharing one reset time. This PR surfaces that second number on the device, only for accounts that actually have it — and generalizes to any per-model weekly bucket the API may serve.

## What it does

The Weekly card **rotates through 1+N faces**, switching in step with the status ticker's word change (~4s): the classic Weekly face (all-models %, usual utilization colors), then one face per weekly scoped-model limit. Each scoped face shows that model's percent with a pill labeled by the **API's own display name**; the bar fills in `THEME_BLUE` when the scoped model is Fable (matching the Claude usage-settings blue) and in the theme grey for any scoped model the firmware doesn't recognize — so if a separately-metered Sonnet/Opus bucket ever returns (the usage endpoint still carries their legacy `seven_day_opus`/`seven_day_sonnet` fields), it rotates in automatically with no firmware change. The swap is instant (no bar-slide animation), the card's geometry never changes, and the reset line is truthful on every face because all weekly limits reset at the same instant (verified against the API — the scoped reset timestamp is byte-identical to the 7d reset). Works unchanged on every board size, 240×240 included.

| Weekly face | Fable face | Unrecognized scoped model (grey fallback, simulated) | Without scoped limits (unchanged) |
| --- | --- | --- | --- |
| weekly face | fable face | scoped fallback | no scoped |

## Conditional by construction

Accounts without scoped weekly limits never rotate and render pixel-identically to today's screen: the daemon omits the `ws` key entirely when the account has none, and the firmware treats key-absence as "feature off" (count 0; `0%` is a legitimate reading and never means "hidden"). Same optional-key pattern as `c` and the Enterprise-only `tp`/`pd`/`rd`. Compatibility both directions: new firmware + old daemon never rotates; old firmware + new daemon ignores the unknown key (ArduinoJson).

## Wire format

`"ws":[{"n":"Fable","p":75}, ...]` — one entry per weekly scoped-model limit, in API order, labeled with the API's display name (fallback to model id). Up to 4 entries are rendered (`MAX_SCOPED_WEEKLY`); names truncate at 15 chars. Well within the 512-byte BLE RX buffer.

When `ws` is present, `w` is re-based on the usage endpoint's `weekly_all` percent instead of the rate-limit header. The two surfaces quantize the same window differently — the header is a 2-decimal fraction (`"0.12"`), the endpoint a rounded integer — so mixing them rendered a real 12.6/12.4 pair as **12/12**, making the card look like it was repeating one number on both faces, and could disagree with the settings UI that shows these same integers. Sharing one source makes the two faces mutually consistent and settings-accurate (verified on hardware: 13% / 12%). Plans with no scoped limit keep the header-derived value untouched, and a failed endpoint lookup falls back to it.

## Design options explored

We iterated through four space-sharing designs before landing on time-sharing. Each was flashed and captured on an AMOLED-2.16 (files live in `screenshots/design-options/`; the folder can be dropped before merge if unwanted).

| 1. Inline mini-row | 2. Labeled rows | 3. Split stripes + legend |
| --- | --- | --- |
| option 1 | option 2 | option 3 |

| 4. Two number rows | 5. Face flipper — **shipped** |
| --- | --- |
| option 4 | option 5 |

1. **Inline mini-row** — thin labeled Fable bar tucked above the reset line, "All models" caption. Too small to read at desk distance.
2. **Labeled rows** — label column left, stacked full-height bars right. Steals bar width; busy.
3. **Split stripes + legend** — the bar splits in half inside its exact envelope, blue Fable stripe below, "Fable NN%" legend on the reset row. Compact, but the stripes are thin.
4. **Two number rows** — the card grows a second big-number row (73% / Fable pill) with a small two-line reset between. Crowded, and squeezes the layout budget on every board.
5. **Face flipper** — **shipped.** Full-size presentation for every number, zero geometry change, trivially conditional, generalizes to N scoped models.

## Where the numbers come from

The scoped percents are deliberately **not** read from the `/v1/messages` rate-limit headers the poll uses. The scoped headers (`anthropic-ratelimit-unified-7d_oi-*`) only appear on responses to requests made *with* the scoped model — polling that way would spend the very allowance being measured, and would fail outright for accounts without access. Instead, all three daemons (macOS/Linux Python, Windows Python, bash) make one extra free `GET https://api.anthropic.com/api/oauth/usage` (the same endpoint `claude /usage` renders, same OAuth token) per poll on Pro/Max accounts, and read every `limits[]` entry with `kind: "weekly_scoped"` and a model scope. No entries → no key. The Enterprise branch never queries it (no weekly window there). Any failure of the extra request omits the key without affecting the main payload.

## Verified

- All six envs compile (`waveshare_amoled_216`, `_18`, `_216_c6`, `_18_c6`, `_206`, `waveshare_lcd_154`).
- On-device (AMOLED-2.16, live daemon): the real two-face rotation (Weekly/Fable) plus a three-face rotation with a daemon-injected fake Sonnet entry — confirming the label-driven pill and grey fallback (third screenshot above). Live transitions between daemons with/without the `ws` key exercised. A pixel-diff of the no-scoped state against pre-change firmware confines all differences to live values (percent digits, bar fill, reset text, status animation) — zero layout drift.
- Daemon suite: 151 passed, 2 skipped (32 new tests in `daemon/tests/test_fable.py` covering present/absent/multiple scoped models, label fallback to model id, Enterprise never querying the endpoint, endpoint failures, 0%-is-real, clamping, the single-source rounding rule and its header fallback, and wire size vs the 512-byte BLE buffer, across both Python daemons). The bash daemon's extraction pipeline was verified against the live endpoint.


nathanjohnpayne:
Hi @HermannBjorgvin, I'm back from vacation (where I made and launched "**[GayCruiseBingo](https://gaycruisebingo.com/)**"), and I kept needing to look up my Fable limit while debugging that app from the ship, so I figured it would be nice to have that in Clawdmeter. I explored a few options; I landed on a meter that changes at the same rate as the flipper at the bottom. Let me know if you like any of them.

HermannBjorgvin:
@jkbdco has been working on https://github.com/HermannBjorgvin/Clawdmeter/pull/128 with the same functionality. I appreciate the design mock ups they are really cool, I had one suggestion that you might like on that PR in a comment here https://github.com/HermannBjorgvin/Clawdmeter/pull/128#issuecomment-5169559992

I am a bit worried about the main screen getting cluttered and my idea there seemed like the most reasonable way to display both the weekly and fable limits together, rotating the screen on a timer might be an option also like you outline. I think the fable limit is really the only model limit anyone is worried about now, I have been anxiously monitoring it myself as well.

HermannBjorgvin:
Closing, on the 14th of September the Fable model will no longer have a separate limit. We can reopen if the policy changes back in the future.

PR #144, opened 2026-08-05
feat: animate the corner creature on the usage screen

TB1982:


TB1982:
Closing this in favour of #153.

#153 brings its own corner mascot, so this one is superseded on the feature itself. It also removes `tools/convert_to_c.js`, which this branch modifies — that is a delete/modify conflict rather than something a rebase would sort out, and the 20×20 pipeline it was built on is not there any more.

No point leaving it in your queue. Thanks for taking a look at it while it was open.

(#151 is unaffected by any of this — it is a daemon-payload fix with no animation code in it.)

PR #145, opened 2026-08-05
Feature/reset celebration

TB1982:


TB1982:
Closing this in favour of #153.

Three of the files this branch touches are removed by #153 — `tools/convert_to_c.js`, `tools/scrape_claudepix.js` and `tools/claudepix_data/_index.json` — so it is a delete/modify conflict rather than a rebase. And the celebration is keyed to `#define SPLASH_CELEBRATE_ANIM "dance djmix"`, which is one of the claudepix animations #153 drops, so even if it merged cleanly it would find no animation and quietly do nothing.

The behaviour is worth having on the official set at some point — a 30 s celebration when a usage window refills — but it would want rebuilding against the new engine rather than carried over. Closing so it is not sitting in your queue as if it were ready.

PR #149, opened 2026-08-08
Add a 20x20 animation editor, the two unlisted claudepix animations, and corrections to six more

TB1982:
## What this adds

**`tools/anim_editor.html`** — an editor for the 20×20 animations. One file, no
dependencies, no server, nothing fetched; open it off disk. It's deliberately
narrower than a general pixel editor: 20×20 and the ten-colour cap are enforced
rather than advisory, and it previews on black at both scales the device
actually uses — 24px/cell on the splash, 4px/cell in the corner badge. Those two
previews are the part a general editor can't give you. A colour that reads on
white can vanish on an AMOLED panel showing black, and a shape that reads at
24px turns to mush at 4px.

Every animation is embedded as a loadable sample, so an existing one can be
opened and fixed rather than rebuilt. Export lands in a new `tools/drawn_anims/`
which `convert_to_c.js` globs — no index to maintain. Four languages
(zh-TW/en/ja/zh-CN), because the animations came from a community and the tool
may as well be usable by one.

**Two animations that were already being scraped and thrown away.** The
manifest carries fifteen entries; the site renders thirteen. `dance bob` and
`work type` were being fetched and discarded.

**Corrections to six of the scraped animations, and four new ones.** Details
below — this is the part worth your judgement, not mine.

## The animation changes

Six replace the scraped original by name rather than sitting alongside it:

| | what was wrong |
|---|---|
| `expression sleep` | the snore bubble was drawn in the body colour — an orange snore |
| `work think` | the same defect one animation over: the thought marks were body-coloured too |
| `dance sway` | floating marks in the top-right, also body-coloured, connected to nothing |
| `idle breathe` | the drifting specks never landed twice in the same place; they're one amber dot now, blinking in place like a cursor waiting |
| `dance bounce dj`, `dance sway dj`, `dance djmix` | his head sat on row 0, so the band arcing over the headphones was outside the 20×20 grid — cropped away entirely, leaving two ear cups with nothing joining them |

The pattern in most of those is one thing: **effects drawn in the body colour**.
On a black panel they read as pixels that fell off him rather than as something
he's doing.

New: `idle hearts`, `idle blossom`, `swim summer`, `work out`.

`dance bob` and `work type` are in `splash_animations.h` but listed in no group,
so nothing picks them. They're the two the source site itself never lists, and
both are four frames with almost nothing moving — `work type` changes exactly
four pixels between frames, with no keyboard and no surface, so there's nothing
in it to read as typing. `swim summer` and `work out` took their slots. The data
is kept; deleting someone else's art isn't ours to do.

**If you'd rather keep the originals and have these sit alongside, say so** —
it's a rename in `tools/drawn_anims/` plus group entries, and I'll push it.

## Licence note

`tools/README.md` has a "License note" that said the scraper hits a site without
a stated licence. That understated it, and this PR adds hand-edited derivatives,
so it's rewritten to say plainly: two parties hold rights here — @amaanbuilds
for the claudepix art and Anthropic for the character — this project holds
neither, and it licenses nothing to anyone, because publishing a derivative
under an open licence would assert a right it doesn't have.

The same disclosure is inside the editor itself, in all four languages, because
that file is self-contained and will be met detached from this repo.

## Scope

27 files. Everything is under `tools/` except three:

- `firmware/src/splash.cpp` — `GROUP_NAMES` and `GROUP_MAX`. Needed: a name
  absent from a group is never picked, so new animations would ship dead.
- `firmware/src/splash_animations.h` — generated by `convert_to_c.js`.
- `CLAUDE.md` — the splash-animation section.

`convert_to_c.js` also validates grid shape, palette size and cell range now.
The firmware indexes these arrays with fixed strides and no bounds checks, so a
19-row grid doesn't render wrong, it reads past the end of an array.

Builds clean on `waveshare_amoled_216` and runs on the hardware.

## Credit

The animations are @amaanbuilds' work from
[claudepix](https://claudepix.vercel.app), as the README already says. The
editor and the corrections are mine.


TB1982:
Pushed one more commit to this branch (`25ae76b`).

The editor bundles the whole catalogue as loadable samples, and the credit line already named claudepix, @amaanbuilds and Anthropic. What it did not say was **why** they are bundled at all — and that omission is what makes the same bundle read as either a teaching aid or a helping of someone else's art taken without asking.

The reason they are there: an empty 20×20 grid is exactly where someone with no pixel-art background stops. The samples are the ramp — something to copy, pull apart and alter until the tool makes sense. Shipping the editor without them would leave a beginner staring at a blank grid, which defeats the point of shipping an editor.

So the note now says that, in all four locales, and adds the part an attribution is worth little without: **if either claudepix or Anthropic would rather the samples were not included, an issue removes them.** If you would prefer the upstream copy ship without them, or with neutral placeholder shapes instead, say so and I will change it — I would rather you did not have to make that call silently.

HermannBjorgvin:
Closing in favor of #153 sorry

PR #152, opened 2026-08-08
Feature/multi size grid

TB1982:


TB1982:
Opened by mistake — I clicked through the banner GitHub shows right after a push. This branch isn't ready for review: it's a format change that lets splash animations carry their own grid size, parked in my fork while the 40x40 render path on the PSRAM-less C6 boards is still unverified. Closing so it doesn't sit in your queue. Sorry for the noise!

PR #154, opened 2026-08-12
feat: OpenCode Go usage on a third screen

gbrein:
Adds an OpenCode screen next to the Anthropic one, with the same two-gauge layout, because the data has the same shape: OpenCode Go exposes rolling / weekly / monthly windows, each with a percent and a reset.

Daemon (daemon/opencode_usage.py, shared by all three daemons):
- Gauges from GET https://opencode.ai/zen/go/v1/usage, authenticated with the key OpenCode already stores in auth.json. Undocumented endpoint, found by probing; the Zen docs list none.
- The User-Agent header is load-bearing: Cloudflare fronts opencode.ai and 403s urllib's default Python-urllib/3.x with error code 1010.
- Today's tokens from a read-only sqlite3 query against opencode.db. Not ccusage: that parses the storage/message/*.json layout current OpenCode versions no longer write, so it silently reports nothing.
- urllib, not httpx, since the bash daemon runs this under the host's python3.
- Every failure path omits the fields, so the screen reads "No data" rather than a fabricated 0%. Free-ride on the stored key, never refreshed.

Firmware:
- Screens now cycle in enum order (splash -> usage -> opencode) instead of the old two-way toggle; prev_non_splash_screen goes away.
- OpenCode's official mark takes the corner slot on its own screen, sized to match clawd_still.h so both marks carry the same visual weight.

Also here, both prerequisites for verifying the above on Windows:
- boards/sim/shim: strlcpy + gmtime_r under _WIN32. Pre-existing gap; the sim had never been built with mingw, which has neither.
- tools/package.json: png_to_lvgl.js needs pngjs and never declared it.
- flash-mac.sh only probed /dev/cu.usbmodem*, missing every USB-UART bridge board (CH9102 enumerates as cu.wchusbserial*).

main.cpp and ui.cpp also carry a few in-flight hunks from the board-port work that interactive staging was unavailable to separate.

Claude-Session: https://claude.ai/code/session_015vSCPLyj5rcQssERdM6iy5


HermannBjorgvin:
Alternative vendors out of scope for the time being sorry.

PR #158, opened 2026-08-25
feat: optional Codex (OpenAI) usage screen fed from local Codex CLI logs

yucapro:
Adds an optional third screen showing OpenAI Codex usage next to Claude's, for desks where both CLIs are in daily use.

Codex screen

## How it works

**Data source (daemon):** Codex CLI writes a `token_count` event into its session logs (`~/.codex/sessions/YYYY/MM/DD/*.jsonl`) after every turn, carrying `rate_limits` (used percent, window minutes, reset epoch) and cumulative token usage. The new `daemon/codex_source.py` reads the newest snapshot plus today's and the prior 7 days' token totals. Purely local file reads: no network calls, no OpenAI auth handling, nothing touches their endpoints. The reset countdown is recomputed from the stored epoch on every poll, so it stays live between Codex sessions; a lapsed window reports 0% rather than stale numbers.

**Payload:** one optional `"x"` object on the existing JSON (`p`/`rm`/`wm` primary window, `p2`/`rm2`/`wm2` secondary when the plan has one, `ti`/`to` today's tokens, `da` 7-day daily average). Old firmware ignores the key, hosts without Codex never send it, and `codex = off` in the config disables it explicitly, so the default experience is unchanged for everyone else.

**Firmware:** `SCREEN_CODEX` joins the tap cycle (Splash -> Usage -> Codex) only while live payloads carry Codex data, and falls back to the Usage view when data goes stale or BLE drops. Panels reuse `make_usage_panel()` and the responsive layout, so all board ports inherit the screen. Slot 1 shows the primary rate window; slot 2 shows the secondary window when the plan has one, otherwise a Daily panel measuring today's total tokens against the 7-day average (bar at 50% = an average day), since Codex exposes no daily rate limit. The corner slot swaps the Clawd mascot for a white OpenAI mark and the bottom line runs the same animated status as the Usage view, in white.

## Testing

- 12 new tests for `codex_source.py` (`daemon/tests/test_codex_source.py`); full daemon suite passes (125 passed, 2 skipped; `test_windows_icon.py` needs PIL and fails on main too).
- Firmware builds clean for `waveshare_amoled_216`, `waveshare_amoled_216_c6`, and `sim`.
- Verified end to end on a Waveshare ESP32-C6-Touch-AMOLED-2.16 with the macOS daemon: pairing, tap cycle, stale-data fallback, and live values against real Codex logs. One-window and two-window plan layouts verified in the SDL2 simulator.
- Windows daemon is wired identically (module is OS-neutral) but only syntax-checked here; the bash Linux daemon is left as a documented follow-up.

## Notes for review

- The corner logo is the OpenAI mark from the simple-icons set (CC0 glyph, but the mark itself is OpenAI's trademark). If you'd rather avoid shipping it, I'm happy to swap in a plain "Codex" wordmark; it's one icon asset either way.
- Key `"x"` was picked to stay terse and neutral; happy to rename if you have a schema preference for future providers.


HermannBjorgvin:
Sorry support for other vendors than Anthropic is currently out of scope for this project.

yucapro:
No worries, totally understandable, thanks for taking a look! For anyone who does want a Codex screen alongside the Claude one, the feature lives on at [yucapro/Clawdmeter](https://github.com/yucapro/Clawdmeter/tree/feature/codex-usage-screen). Closing this. Thanks for the great project!

PR #159, opened 2026-08-25
Feature/daemon token renew

TB1982:


TB1982:
Sorry for the noise — opened against this repo by mistake. This change is fork-specific (it nudges Claude Code to refresh an expired OAuth token so the device stops showing "No data" when nobody happens to be running a session), and it was only ever meant for my own fork. Closing.

PR #160, opened 2026-08-30
fix(c6): restore correct orientation for AMOLED-2.16 C6

JKutac:
## Problem

On the Waveshare ESP32-C6-Touch-AMOLED-2.16 (`waveshare_amoled_216_c6` env), the display renders rotated 90°/270° from correct after flashing current `main`.

## Root cause

9375bd6 ("fix(c6): drive AMOLED-2.16 C6 panel as CO5300, side-USB orientation") correctly rebased this board's panel driver from `Arduino_SH8601` to `Arduino_CO5300` and fixed the panel-driving-voltage registers — that part is right and is kept as-is.

However, the same commit also dropped a raw `MADCTL 0x30` (MV transpose + ML) write that the prior version relied on for correct orientation, switching to the CO5300 class's default `MADCTL 0x00` under the assumption that this was "the preferred desk orientation" (USB port on the side). That assumption doesn't hold — it was never actually verified visually on hardware, only touch calibration was re-tap-tested to match whatever the display was doing.

## Fix

- `display.cpp`: restore the `writeC8D8(0x36, 0x30)` MADCTL write (MV transpose + ML) after the panel-driving-voltage registers, matching what the pre-CO5300-rebase version wrote.
- `touch.cpp`: revert touch axis calibration to `setSwapXY(true)` / `setMirrorXY(true, false)` to match the restored display orientation.

This is a minimal, targeted partial revert of 9375bd6 — only the orientation-related lines change; the CO5300 driver rebase and panel-driving-voltage fix are untouched.

## Verification

Built and flashed to a physical Waveshare ESP32-C6-Touch-AMOLED-2.16. Confirmed the display now renders right-side-up (previously sideways) and touch input tracks correctly at the corrected orientation.

### DIFF
--- a/firmware/src/boards/waveshare_amoled_216_c6/display.cpp
+++ b/firmware/src/boards/waveshare_amoled_216_c6/display.cpp
@@ -30,8 +30,8 @@ void display_hal_init(void) {
 // does not set is this panel's manufacturer page-0x20 driving-voltage
 // registers (0x19/0x1C) — without them the panel stays black even with the
 // rails up. Set just those; everything else the SH8601-era hack also wrote
-// (0xC4/0x36/0x53/0x51/0x63/0x29) is now covered by the class init.
-//
+// (0xC4/0x53/0x51/0x63/0x29) is now covered by the class init, and we override
+// MADCTL below to fix orientation.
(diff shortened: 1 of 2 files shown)


HermannBjorgvin:
Thanks for the PR!

PR #164, opened 2026-09-07
History and Maxing screens: usage over time, from local transcripts

tcbutler:
## What this adds

Two screens joining the tap cycle (splash → usage → **history** → **maxing** → splash), answering "how much have I been using, and when do I run out?" — questions the current single-window view can't.

Both are fed from the Claude Code transcripts already on disk (`<config_dir>/projects/*/*.jsonl`). **No extra API access and no extra tokens.**

**History** — 14 daily bars of output tokens with the current week's running total drawn over them, then Today (with turns and % vs the 7-day average), This week, and a Pace projection: *"62% at reset"* / *"Limit in 1h 20m"* / *"Idle"*, computed on-device from the existing burn-rate ring buffer. Plus a model-family mix bar.

**Maxing** — a week of 5-hour limit windows, one row per day, five fixed time-of-day columns (00-05 … 20-24), shaded none / some / most / maxed. It answers *when* you max out, not just how often.

## The part worth reviewing carefully

5-hour windows are **reconstructed** from the transcripts: the window is first-use anchored and runs exactly 5 h, so walking the turn stream rebuilds the boundaries. Validated against a live account — the reconstructed open time matched the one implied by the reset header.

But **utilisation has no ground truth for past windows**: the API reports a percentage for the *current* window only. So every cell is one of three things, and the payload encodes which so the screen never presents inference as fact:

| | meaning | on screen |
|---|---|---|
| lowercase `a`-`d` | measured, and watched to the window's close | no border — settled fact |
| uppercase `A`-`D` | measured, but the daemon stopped watching early | grey border — "at least this much" |
| digits `0`-`3` | estimated from token volume | grey border |

The estimate divides output tokens by a tokens-per-percent ratio **learned from the account's own measured windows**, not a hardcoded constant. The greys thin out on their own as windows get watched.

**Two limitations I'd rather state than have you find:**

1. **Cloud, mobile and other-machine usage is invisible to these screens.** It counts against the limits, so the usage screen's percentages stay correct, but it isn't in the local transcripts. Verified: every turn carries `entrypoint` of `claude-desktop` or `cli`.
2. **macOS daemon only.** The Windows and bash daemons are untouched and simply send no history fields; the firmware ignores their absence and the screens show "No history yet".

Worth noting Claude Code's own `/usage` does the same thing for the same reason — *"computed from local session history on this machine"* ([docs](https://code.claude.com/docs/en/costs)). There is no first-party historical endpoint for subscription plans; the Admin API is explicitly unavailable to individual accounts.

## Relationship to prior work

This is deliberately **not** another attempt at the session/context-window screens in #129, #134 and #136 — different data, no hook integration, no daemon-side session watching. It follows your steer on #128: *"instead of fitting everything into one screen just add another view you can switch to by tapping"*, and *"it has to look purposeful"*.

No per-model weekly bucket is surfaced. Nothing in the firmware consumes it, the payload is already near its ceiling, and Fable's separate weekly limit goes away on 14 September — the reason you closed #128 and #143.

## Constraints hit along the way

**LVGL's heap is what limits widgets here, not board DRAM.** The pool is a fixed 64 KB `.bss` array. The first cut of the Maxing grid used 56 cells with per-object *local* styles and exhausted it — `splash_init()` then hung inside `lv_array_resize`. Cells and history bars now use **shared `lv_style_t` objects**: 46.6 KB used, ~10 KB free. Two rules follow, both recorded in CLAUDE.md: prefer a shared style on repeated objects, and never `lv_obj_remove_style(obj, NULL, part)` to swap one, because in LVGL 9 `set_size`/`set_pos` live in the object's local style and that wipes geometry too.

**Payloads with history exceed a write-without-response** (ATT_MTU-3; CoreBluetooth truncates silently). The daemon now sends anything over 160 bytes as a GATT long write, bounded by a timeout. The RX characteristic already allowed both. Payloads under the threshold keep the old path byte-for-byte.

## Daemon changes

- `daemon/usage_history.py` — incremental transcript aggregation. Per-file offsets persisted, so the cold scan is the only expensive one: **923 files / 1.28 GB in ~2 s**, then milliseconds. Versioned state file, because offsets sit at EOF and a newly-derived field would otherwise stay empty forever on an existing install.
- **Primary polling moved to `GET /api/oauth/usage`** — the endpoint Claude Code's `/usage` reads, zero tokens, exact reset timestamps. The 1-token `/v1/messages` path remains the fallback for a non-200, malformed JSON, or any non-Pro/Max shape, which keeps Enterprise detection exactly where it was. **A 429 benches it for 15 minutes**; I read #29/#37 and did not remove that.
- **Polling no longer requires a connected device.** It used to run only inside the BLE loop, so unplugging the board left permanent holes in the history and calibration. Only safe to change because polls are now free.

`history = off` in the config skips all of it.

## Testing

- **306 daemon tests** (`pytest daemon/tests`), 14 host tests on the pure layout/pace maths (`firmware/test/test_history_math/`).
- All nine build targets green: AMOLED-2.16 S3/C6, AMOLED-1.8 S3/C6, AMOLED-2.06, LCD-1.54, LCD-4, sim, sim_cyd.
- Verified on hardware over a live daemon: 325-byte payloads delivered every poll, zero parse errors, no resets. Polling-while-disconnected confirmed by holding the board offline for 200 s and watching five history writes land.
- Adds a **`sim_cyd`** env — the desktop simulator at 240x320 — since small-screen layouts otherwise can't be checked without a PSRAM-less board's screenshot path.

Happy to split this further or drop either screen if you'd rather take them separately.


HermannBjorgvin:
Hard to review, no screenshots of the concept and I'm not reading more than the first paragraphs because I don't like reading the writing coming from the Anthropic models these days.

Could you attach some screenshots? Also why 14 bars why not 1 per day of the week?

Also have you considered reusing an existing package like ccusage? I would really like to avoid the overhead of having to maintain some usage analysis of CC usage. The Daemon could possibly look to see if it's installed or use npx to use ccusage.

The 5 hour window thing I'm not really getting, fixed 5 hour windows doesn't really make sense since they don't all start at the same time for people.

Please explain, ideally handwritten and not an AI summary.

PR #165, opened 2026-09-08
sim: build with -std=gnu++17 for macOS

igrkov-droid:
On macOS `g++` is a wrapper around Apple clang, which defaults to C++98 when invoked under that name (__cplusplus reports 199711L). The sim env never set -std, so the build died on constexpr and brace-init in both ArduinoJson and usage_rate.cpp. Linux `g++` is real GCC and defaults to gnu++17, which is why this only bites on macOS.

### DIFF
--- a/firmware/platformio.ini
+++ b/firmware/platformio.ini
@@ -476,6 +476,10 @@ build_src_filter =
 
 build_flags =
     -DBOARD_SIM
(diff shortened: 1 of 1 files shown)


HermannBjorgvin:
Thanks for the PR!

PR #168, opened 2026-09-15
Physical Deny/Always/Allow approval for Claude Code PreToolUse prompts

matth-production:
## Summary
- New PERM_REQ/PERM_RESP BLE characteristics (`...0005`/`...0006`) so the host can relay a pending Claude Code tool-call approval to the device, which shows a dedicated permission screen (tool name + short summary) with Deny/Always/Allow as touch targets, and notifies back the tapped decision.
- `daemon/permission_hook.py`: a Claude Code `PreToolUse` hook that talks to a new Unix socket (`~/.config/claude-usage-monitor/permission.sock`) on the existing macOS usage daemon — reusing its already-open BLE connection rather than each tool call opening its own. Fails to `ask` (never a silent allow/deny) on any hardware problem — daemon not running, device not connected, nobody tapped in time — so a hardware hiccup just falls back to the normal in-terminal prompt, it never blocks or rubber-stamps a tool call.
- **"Always Allow" persists** two ways: (1) a narrow, exact-match rule in `.claude/settings.local.json`'s `permissions.allow` — the same file/array Claude Code's own "don't ask again" writes to — that survives across sessions but only matches an exact repeat; and (2) a **session-scoped quiet** for the whole program (e.g. `python3`) that lasts only for the current Claude Code session, so a busy session calling the same program with different arguments each time doesn't keep re-prompting, without permanently broadening what's allowed.
- **Never prompts at all** for a set of commands that are safe regardless of arguments — `grep`/`cat`/`ls`/`head`/`tail`/`wc`/`pwd`/`echo`/`which`/`whoami`/`file`/`stat`/`du`/`df`/`date`/`cd`/`mkdir`, plus read-only git subcommands (`status`/`diff`/`log`/`show`/`branch`) — checked per-subcommand across compound/piped commands. Deliberately excludes `python`/`python3` and anything else capable of arbitrary action.
- Simplifies the usage screen's status line: dropped the spinner glyph + whimsical word-cycling (`"Calculating…"`, `"Frolicking…"`, etc.), which forced an LVGL label redraw roughly every 130ms for purely decorative reasons. Kept the functional states (Waiting/Connected/Listening/No data), now only redrawn on an actual state change.
- Forces a display wake (`idle_note_activity()`) when a permission request arrives — otherwise a time-sensitive prompt could render correctly behind a dimmed/sleeping panel and look like nothing happened.
- **Enables audio on the ESP32-C6 AMOLED-2.16 board** (previously a no-op stub) — ES8311 codec + I2S pins verified against the official Waveshare XiaoZhi BSP for this exact board, no PA-enable GPIO needed on this variant. Volume tuned to 85 (max clips audibly, no separate analog gain stage on this board). Also corrects a `board.h` comment that wrongly assumed a TCA9554 IO expander was involved in the audio path.
- **Chimes on an incoming permission-approval request** so it's actually noticed, not just silently rendered.
- Two new entries in `CLAUDE.md`'s gotchas list: macOS/CoreBluetooth caching a bonded peripheral's GATT table across a characteristic-set change (needs a Bluetooth toggle or re-pair once), and BLE-triggered screens needing an explicit wake call.
- README gets a new "Permission approval on the device" section with setup instructions and known limitations.

Built and tested end-to-end on a Waveshare ESP32-C6-Touch-AMOLED-2.16, including registering the hook in a real `settings.json`, confirming Bash/Write/Edit/NotebookEdit tool calls route to the device instead of the terminal prompt, confirming the chime is audible on real hardware, confirming the Always-Allow persistence loop (tap → rule lands in `settings.local.json` → an exact repeat auto-allows with zero device round trip), confirming the safe-command list never reaches the device, and confirming session-quiet correctly scopes to one session only.

## Known limitations (documented in the README)
- Persisted (permanent) rules are scoped per-project, matching Claude Code's own native behavior.
- Session-quiet state is a small file per `session_id` under `~/.config/claude-usage-monitor/session_quiet/` — never explicitly cleaned up, just inert once a session ends (negligible size/lifetime for a personal device).
- One physical screen — concurrent tool calls serialize through the pending request rather than each getting their own prompt. If you run multiple Claude Code sessions at once, they all share the one physical screen (the hook is registered globally in `settings.json`, not per-project).
- The Linux bash daemon (`claude-usage-daemon.sh`) doesn't implement the PERM_REQ/PERM_RESP side yet — this is macOS-only for now.

## Test plan
- [x] Firmware builds clean for `waveshare_amoled_216_c6`
- [x] Flashed to real hardware; permission screen renders and buttons respond to touch
- [x] Full round trip verified: hook → daemon socket → BLE write → device screen → tap → BLE notify → daemon → hook stdout, for Allow, Always, and Deny
- [x] Registered as a real `PreToolUse` hook in `settings.json` and confirmed live Bash tool calls route to the device instead of the terminal prompt
- [x] Audio: ES8311/I2S init confirmed via serial log (`chime: ES8311 ready`), chime audibly confirmed on real hardware, including firing alongside an incoming permission request
- [x] Always-Allow persistence: rule confirmed written to `.claude/settings.local.json`; exact repeat of the same command auto-allowed with no daemon log entry and no device round trip
- [x] Safe-command list: `grep`/`cd`/`mkdir`/etc. confirmed to never generate a daemon log entry at all
- [x] Session-quiet: confirmed in isolation (different session_ids don't leak into each other; same session_id + same program key correctly matches across different exact commands)
- [ ] Not tested on other boards (S3 2.16, 1.8, 2.06, LCD variants) — the permission-approval work only touches shared `ui.cpp`/`ble.cpp`/`main.cpp`, no board-specific files changed there, but worth a look on hardware you have handy. The audio commit is C6-2.16-specific only.
- [ ] Linux daemon parity not implemented


HermannBjorgvin:
Sorry this is out of scope for now I think. Auto mode mostly solves the permission prompts I don't know who is still manually approving tool uses.

PR #170, opened 2026-09-15
usage: show a spent limit instead of "no data"

igrkov-droid:
Running out of quota looks exactly like losing the link: the screen falls back to the idle state and says nothing. That is the one moment the display has something worth saying, and the information is available the whole time — it is discarded on two separate floors.

daemon: any 4xx was treated as a failed poll and returned nothing. But a refusal still carries the rate-limit headers, and when the limit is what caused the refusal those headers hold precisely what the user wants: spent, and when it comes back. Bail only when the response carries no quota headers at all.

firmware: the "st" field was parsed into UsageData and never read — zero references in ui.cpp. It now drives the reset line ("Spent - back in 2h 11m", in red) and the status line ("Limit reached" in place of the idle whimsy, which is for when there is nothing to report).

Verified in the simulator with a payload of s=100, st="limited"; both the sim and the AMOLED-2.16 target build.


HermannBjorgvin:
See #166 , could use some feedback on that indeed.

PR #171, opened 2026-09-15
daemon: let the daemon hold a long-lived token of its own

igrkov-droid:
The daemon reads the OAuth token the CLI keeps, and that token lasts hours: it is refreshed when the CLI is used, not on a schedule. A desk device left alone therefore goes blind — here, twice in four days, each time showing "No data" with nothing on screen tying it to an expired credential.

`claude setup-token` exists for this and mints a year-long token. This accepts it from either place: $CLAUDE_CODE_OAUTH_TOKEN, which is what that command suggests and what a launchd or systemd unit can carry, or ~/.config/claude-usage-monitor/token for a plain install. Either wins over the CLI's own token; otherwise nothing changes.

Warns when the file is group- or world-readable, and logs which source is in use — a token file that is present but ignored (wrong path, empty after a botched paste) is otherwise indistinguishable from everything working, until the screen goes blank hours later.

Deliberately not implemented: refreshing the CLI's token via its refresh token. That means writing back to the record the CLI owns, and servers commonly rotate the refresh token on use — getting it wrong costs the user their Claude Code login. Not a trade a background daemon should make.


HermannBjorgvin:
A good idea

PR #172, opened 2026-09-17
C6 AMOLED-2.16: IMU auto-rotation via CO5300 MADCTL (no rotation strip)

pavlovicnemanja:
## What

Enables IMU-driven auto-rotation on the **Waveshare ESP32-C6-Touch-AMOLED-2.16** port.

The C6 port had `BOARD_HAS_ROTATION 0` because the S3 approach (CPU rotation strip in `display_hal_draw_bitmap`) needs PSRAM the C6 doesn't have. The CO5300 can rotate on its own through the **MADCTL** register — the port already writes `0x36 = 0x30` for the base orientation (#160) — so this maps each IMU quadrant to a MADCTL value and lets LVGL redraw. No rotation buffer, no extra RAM, `main.cpp`/`ui.cpp`/`splash.cpp` untouched.

## Changes (all under `firmware/src/boards/waveshare_amoled_216_c6/`)

- `imu.cpp` — same poll/hysteresis logic as the S3 AMOLED-2.16 port (10 Hz, 300 ms stable, 30° tilt threshold), plus a 25 % axis dead zone so a unit on a tilted desk stand (gravity split across X/Y) doesn't flap between quadrants.
- `display.cpp` — `display_hal_tick()` now does blank → MADCTL → `lv_obj_invalidate()` → brightness ramp (same feel as the S3 port). MADCTL table anchored on the hardware-measured upright orientation (gravity on +Y = quadrant 3 → `0x30`), 180° = `0xF0`, sides = `0x50` / `0x90`.
- `board.h` — `BOARD_HAS_ROTATION 1`.

## Tested

On the ESP32-C6-Touch-AMOLED-2.16 (Waveshare), `waveshare_amoled_216_c6` env: upright on the stand, both 90° sides and upside down all render upright; transitions blank and ramp back; splash animations and the usage view redraw correctly; BLE/daemon unaffected. RAM 30.8 % / flash 27.2 %.

Touch is left as-is (tap-anywhere toggles views, so it doesn't need remapping) — same as the S3 port today; if #120 lands, the C6 can use the same quadrant-based mapping.


HermannBjorgvin:
Thanks for the PR!

